PR #12, opened 2024-10-19
Handling virtualkeyboard placement

johannesmutter:
So after trying like two dozen different techniques I finally found a solution that works well in Chromium (Android) and semi-ok in Safari. In Safari there are temporary placment issues of the toolbar when the viewport shifts (the user scrolls, but the browser doesn't actually scroll the page, but instead moves the visual viewport up / down until the "edge" is reached" ... hard to explain in words … but it's a known issue and there are several bug reports for Safari. The solution for chromium is just a line of css and an additional meta viewport tag attribute.

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -3,7 +3,14 @@
 	<head>
 		<meta charset="utf-8" />
 		<link rel="icon" href="%sveltekit.assets%/favicon.png" />
-		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		
+		<!-- Challenge: Show the toolbar when the virtual keyboard is visible.-->
+		<!-- Solution: enable the VirtualKeyboard: overlaysContent property -->
+		<!-- Caveat: This only works in Chrome (Android) so far -->
+		<meta 
+			name="viewport" 
+			content="width=device-width, initial-scale=1.0, viewport-fit=cover, interactive-widget=resizes-content"
+		>
 		%sveltekit.head%
 	</head>
 	<body data-sveltekit-preload-data="hover" spellcheck="false">
--- a/src/lib/ListBlock.svelte
+++ b/src/lib/ListBlock.svelte
@@ -11,6 +11,7 @@
   data-type="block"
   data-index={path.at(-1)}
   style="anchor-name: --{path.join('-')};"
+  class='p-10 max-w-screen-md mx-auto'
 >
   <Container class="list" path={[...path, 'items']}>
     <!-- NOTE: We only allow list items inside list  -->
--- a/src/lib/StoryBlock.svelte
+++ b/src/lib/StoryBlock.svelte
@@ -11,13 +11,16 @@
 </script>
 
 <div
-  class="story-block layout-{block.layout} p-2"
+  class="story-block layout-{block.layout} max-w-screen-lg mx-auto w-full"
   data-path={path.join('.')}
   data-type="block"
   data-index={path.at(-1)}
   style="anchor-name: --{path.join('-')};"
 >
-  <div class='non-text-content' contenteditable="false">
+  <div
+    class='non-text-content' 
+    contenteditable="false"
+  >
     <!-- svelte-ignore a11y_img_redundant_alt -->
     <img src={block.image} alt="Random image" />
   </div>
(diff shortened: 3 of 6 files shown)


johannesmutter:
How does everyone else solve this?

Seems they don't …

For example Notion/ Fibery/ LinkedIn/ ... have their toolbar fixed over the keyboard in their native apps, but floating somewhere on the screen near the selection in their web apps.

This works for them since the blocks one select here are usually just a single line of text, and not a block that spans multiple  rows/ an area beyond the visible screen.

CleanShot 2024-10-19 at 13 04 28@2x

image


image

johannesmutter:
Another insight: regardless of the placement of the toolbar we need to manually control wether the virtual keyboard should be shown or hidden, otherwise click on any tool will cause the toolbar to disappear as the focus is no longer on a "text" node.

There's a new API for this:
[VirtualKeyboard: show() and hid() method](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard)
(No support on Safari yet).

michael:
We may be able to keep the selection when clicking the tools, if we use preventdefault. However, we need to careful that this doesn’t mess with other behavior.

rgon:
The placement of the keyboard can be elegantly solved with pure CSS by using the `dvh` unit [('dynamic' viewport height)](https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic). It represents the 'dynamic height', so it will also resize with all the on-scroll navbar resize events, but that's actually desired here, so you may safely ignore the [common warnings](https://dev.to/frehner/css-vh-dvh-lvh-svh-and-vw-units-27k4).

I suppose all UI components can be wrapped inside an 'overlays' div with `height: 100dvh` with `pointer-events:none` to make it transparent to interactions. Or if that's not desirable, `.editor-toolbar` could maybe be positioned as `top: 90dvh` or maybe `top: calc(100dvh - var(--s-4));`, that's the quickest way to test that it works.

Just keep in mind that it's not supported in Firefox [(caniuse)](https://caniuse.com/?search=dvh), so a fallback would be required:
```css
top: calc(100vh - var(--s-4));
top: calc(100dvh - var(--s-4));
bottom: auto;
```
or if you go with the `.overlays` container option:
```css
height: 100vh;
height: 100dvh;
```
However, since this only happens in mobile, and Firefox does resize the canvas to the remaining space if the keyboard is open, `100vh` will in fact be the correct dimensions, so this fallback will work as expected (source: the [link](https://www.youtube.com/watch?v=xl9R8aTOW_I&t=1825s) @johannesmutter posted)

rgon:
> click on any tool will cause the keyboard to disappear (toolbar moves) as the focus is no longer on a "text" node.

This can be fixed by calling `.preventDefault()` on the toolbar's button `mousedown`: see [REPL](https://svelte.dev/playground/hello-world?version=5.0.5#H4sIAAAAAAAAA52RMU_EMAyF_4rJUpDQ3V7aSki3MLCxUYa09ekiUrtKnBZU9b-TNCdxAxNDJOfF-t6zs6qzsehV-b4q-Z5QlbugHhXpMd2ep-ngZ7SStE57_EvvmQRJIkZVvndmkqalViwKpHaooVjY2aF4aqk6_nZQZWgKAsm5bpXgl7QKZm1Dunq02AuMeAcvsDAVApY9wpn74FsVAdVgZria42BEd3bnuIDx_VV_ImjIGMMEhuCCLmo0gFyQYHLofSqhCyIcs0VgDta55vZkr32qKvcCUzly8DjwQvUah2wAD5E4xzAnPOtg5f5ha96Ybafd1cFXx1z8B3ViIBYYMI90g8rB40-kFaoyLWD72H4An2Nsit0BAAA=)

I only tested on Safari, IOS 18, my Pixel broke yesterday.

johannesmutter:
These properties may help with the iOS context menu appearing for text selections:
```
    -webkit-user-select: none;
    -webkit-touch-callout: none;
```

michael:
Sorry, will reopen it. Just thought that we need generally a different approach to this.

johannesmutter:
Relevant docs for fixing the virtual keyboard resize issues:

https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API

https://developer.mozilla.org/en-US/docs/Web/CSS/env#keyboard-inset-top

https://www.htmhell.dev/adventcalendar/2024/4/

https://interop-2022-viewport.netlify.app/combined/viewport-units/

PR #68, opened 2025-08-12
Implement automatic cursor trap orientation using CSS container queries.

johannesmutter:
Developers can now provide hints in their components via a css variable wether the layout is row or column oriented. And change this variable also depending on screen resolutions or other factors with media/ container/ style queries. No javascript necessary.

- Replace JavaScript-based orientation detection with pure CSS solution
- Add container query support for --layout-orientation custom property
- Simplify CursorTrap component by removing node_array_path dependency
- Update ImageGrid, Story, and List components with layout orientation hints
- Remove determine_node_array_orientation utility function
- Improve empty node array cursor styling with responsive sizes
- Update Overlays component to use CSS container queries for cursor positioning
- Remove hardcoded node_types_with_horizontal_node_arrays configuration
- Improved design & aesthetics of rendering cursor traps and blinking cursor

Breaking changes:
- CursorTrap no longer accepts node_array_path prop (no longer required)
- Document config no longer needs node_types_with_horizontal_node_arrays

Developer experience improvements:
- Components automatically detect layout orientation via CSS
- Better visual feedback for empty node arrays
- More responsive cursor sizing based on device capabilities

### DIFF
--- a/src/routes/components/Story.svelte
+++ b/src/routes/components/Story.svelte
@@ -46,19 +46,25 @@
 		gap: var(--s-10);
 		padding: var(--s-10) var(--s-6);
 	}
+	.caption {
+		container-type: inline-size;
+	}
 
 	.story :global(.buttons:not(.empty)) {
 		--column-gap: var(--s-2);
 		--row-gap: var(--s-2);
 		padding-top: 24px;
+	}
+	.story :global(.buttons) {
 		display: flex;
 		flex-direction: row;
 		flex-wrap: wrap;
 		column-gap: var(--column-gap);
 		row-gap: var(--row-gap);
-	}
-	.story :global(.buttons) {
 		--layout-orientation: horizontal;
+		@container (max-width: 320px) {
+			--layout-orientation: vertical;
+		}
 	}
 
 	.story :global(.buttons.empty .node.empty-node-array) {


johannesmutter:
TODOS:
- [ ] CursorTraps in empty NodeArray need better visibility.
- [ ] CursorTraps in NodeArray with no gaps need better visibility, or make sure there's always at least the gap variable specified.

johannesmutter:
> there's a horizontal scrollbar now (which wasn't there before)

It wasn't noticeable before because the padding was larger than the cursor trap. 
Cursor traps where to small to be easily accessible, even with fine pointer devices.
So whenever the padding is smaller than the cursor trap we'll get an overflow. 
We can't just do overflow: hidden, since it will then not render the cursor-trap anymore.
So we need to either enforce that there's always a minimum padding (but that can be too opionoated) or we need to adjust the placement based on the available space. Which is a bit tricky, but I think doable.

CleanShot 2025-08-13 at 02 08 17

johannesmutter:
> i don't really like the mouseover effect (solid line instead of dashed, can we disable it?) i think the pointer is enough, makes the user interface more calm as you move around with the mouse. but that's something anyone can decide in their app code, to spice it up /highlight it on hover.

I understand the preference for a calmer UI without the hover effect. It's why I also reduced the visual noise of the cursor traps to a single line. Also because a cursor trap is really conceptually a line and not a box, mimicking the shape of the text caret.

The hover effect (solid line) can definitely be disabled if you prefer just the pointer cursor. 

I added the hover effect primarily for usability in dense layouts. Since we have very little space between interactive elements (like the image property and cursor traps), the visual feedback helps users target the right element and avoid misclicks. But I totally get that it might feel too busy for some use cases! However I also think it's a good default and the current implementation is subtle. I made a small change and added a tiny delay to avoid flickering when quickly hovering multiple nodes.

The pointer cursor change (on CustomProperty) was definitely needed though - having a text cursor on non-text elements felt wrong, and now the cursor behaviour is more consistent.

Feel free to remove the hover styles if they don't fit your design vision for Svedit! 

It might be worth documenting different styling approaches in a recipes/examples folder for developers who want more opinionated defaults. If you prefer to remove it, can you set this up please?

| Without hover | With hover |
|---|---|
| CleanShot 2025-08-13 at 03 20 24 |  CleanShot 2025-08-13 at 03 25 00  |

PR #70, opened 2025-08-13
fix: exclude previous node when dragging forward from after-node-cursor-trap

johannesmutter:
When dragging forward from an after-node-cursor-trap to the next node, the selection now correctly excludes the node that owns the cursor trap. Preserves backward drag behavior and cross-level selection expansion.

Below video illustrates the before/ after behaviour.
Before a drag selection to the next node would also select the previous node, because the cursor trap belonged to that node.

CleanShot 2025-08-13 at 17 27 04

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -480,15 +480,20 @@
     let focus_root_path = focus_root.dataset.path.split('.');
     let anchor_root_path = anchor_root.dataset.path.split('.');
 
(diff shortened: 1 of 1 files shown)


michael:
Unless I overlook something, the edge-case check can be much simpler, and you can check at the very end:

If you detected a non-collapsed node selection, and the dom selection anchor node is inside an after-node trap, then the computed anchor_index can be decreased by 1.

PR #82, opened 2025-08-19
[WIP] Support for composition events

michael:


michael:
Looking quite good now.

Remaining issues:
- Kicking off composition with a dead key and then hitting ESC breaks things
- When triggering the character selection via long press (e.g. typing and holding "u") and then immediately selecting via number selection (press 2 for the second option), there's no compositionstart event, so we miss that character composition. I have no idea how to fix it other than heuristics in a state machine (like did someone longpress aou followed by typing a number)

PR #85, opened 2025-08-22
[WIP] Use a plaintext-clipboard payload to enable Safari compatibility

michael:


michael:
I use a HTML clipboard now. Closing.

PR #96, opened 2025-08-27
Support Korean&Japanese character/word composition

michael:


michael:
Actually, won't merge this yet, so we have more "solutions" to try/compare side by side. New PR coming with a "dumb" approach to recognize all weird inputs in a single code path.

PR #155, opened 2025-10-25
Prevent flickering of cursor jumping from end of placeholder string to start of text property; Fix disappearing caret in Safari

johannesmutter:
Video with before and after behaviour of this fix. 
Note how the cursor is first placed at the end of the placeholder, and then jumps to the front

https://github.com/user-attachments/assets/c897c7b9-b368-4ef4-a450-0ba5e39a0200

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,9 @@ a simplified version of the markup of `<NodeCursorTrap>` and why it is implement
       Adding a &ZeroWidthSpace; (or any character) here will lead to 2 cursor
       positions (one before, and one after the character)
 
+      Using <wbr> will make it only addressable for ArrowLeft and ArrowRight, but not ArrowUp and ArrowDown.
+      And using <span></span> will not make it addressable at all.
+
       Svedit uses this behavior for node-cursor-traps, and when an
       <AnnotatedTextProperty> is empty.
     -->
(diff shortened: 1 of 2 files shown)


michael:
Unfortunately this breaks multiline placeholders, and doesn't work with text-align: center, as you change the flow of the content.

image

I was trying a lot of different things for the placeholders, and settled with this as the flicker is the lesser evil. 

The flicker only exists in Chrome (Safari has other issues though, like sometimes the cursor disappears visually when the placeholder is active), so maybe I can report this and we have that fixed at the source?

Could you create an isolated reproduction (they need a HTML file, no frameworks) and post it to https://github.com/michael/web-editing/issues, then I will add it to the bug list I'm plannign to submit to the W3C editing group.

michael:
The flicker occurs because we are switching styles on focus to move the cursor into the right place. The original non-flicker version would display the cursor at the end of the placeholder.

So I'm afraid we won't have much like with the browser vendors here. Still interesting that Safari doesn't flicker. Who knows... let's report it.

johannesmutter:
> Can we some apply this rule below only for Chrome?
> 
> ```css
>   [placeholder].empty.focused br {
>     display:none;
>   }
> ```

With CSS: not really longterm reliably, as Safari continuous to add missing features that only chrome has. 
The `@support` query I use works currently. But not sure for how long. A more stable solution would be adding a class ".chrome" with javascript to the `<body>` tag when Chrome is dedected.

Issue #156, opened 2025-10-25
Implement FocusSelectableCommand('previous|next')


johannesmutter:
### Expected behaviour: 
jump to next / previous focusable text property.

### what's actually happening:
focus jumps to other focusable html elements outside of svedit doc, e.g. in demo to "GitHub Stars" button then "Browser Addressbar".

johannesmutter:
IMHO just because contenteditable or prosemirror have no concept for Tab key behaviour / are treated as a single focusable thing, doesn't mean it's a behaviour that should be ignored. Ask yourself what would your or another user expect when pressing the tab key when there's a bunch of fields in front of them, very much visually and from behaviour similar to a regular html form.

Made a PR here with more detailed reasoning why I think this should be core:
https://github.com/michael/svedit/pull/158

johannesmutter:
Good points, haven't thought about Tab to indent a list. 
Establishing a commands list with keyboard mapping sounds like a very developer friendly solution to me 👍

PR #158, opened 2025-10-25
Tab and Shift+Tab key support for fast keyboard driven navigation through a doc and to improve accessibility and integration;

johannesmutter:
I think Tab/Shift+Tab should be handled in core for three reasons:

1. **Accessibility and page integration**: On real pages, Svedit is one stop in the global tab order. If Tab isn’t handled inside the document, all internal fields become a focus “void” and users jump straight from Svedit to the next page control (or the address bar). Keeping Tab within the document (while escaping at the first/last field) preserves a predictable, accessible focus path and makes Svedit feel integrated with the host UI.

2. **Speed and productivity**: Tab lets you jump property-to-property (text and custom properties) and across empty node arrays. That’s dramatically faster than arrow keys, which must traverse every line of text and cursor trap. In practice, you can move through a document up to 10x faster.

3. **Correctness and maintainability**: Determining the next/previous logical field depends on Svedit internals (data-paths, cursor traps, empty node arrays). Pushing this to app-land duplicates brittle logic and will diverge. Implementing it once in core keeps behavior consistent across apps.

Proposed behavior:
- Tab/Shift+Tab moves between logical fields in DOM order: `[data-type=text]`, `[data-type=property]`, and empty node-array placeholders; at boundaries, Tab escapes to the page’s next/previous focus target.

This keeps Svedit composable on complex pages without trapping focus, while making keyboard editing significantly faster.

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -892,7 +892,64 @@ ${fallback_html}`;
       doc.select_parent();
       e.preventDefault();
       e.stopPropagation();
+    } else if (e.key === 'Tab' && !e.metaKey && !e.ctrlKey) {
+      console.log('DEBUG: Tab / Shift+Tab');
+      // Tab / Shift+Tab: Navigate between editable fields (text, custom properties, empty node arrays, but not cursor traps)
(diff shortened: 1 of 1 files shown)


johannesmutter:
Related Discussion:
https://github.com/michael/svedit/issues/156#issuecomment-3447736864

PR #209, opened 2026-02-24
Svedit Node Cursor

johannesmutter:
This continues the work from my [PR #68](https://github.com/michael/svedit/pull/68), which replaced the JavaScript-based orientation detection with a CSS `--is-horizontal` variable and improved cursor trap aesthetics. That PR was closed after Michael extracted the minimal orientation-variable approach into [PR #74](https://github.com/michael/svedit/pull/74) (merged) and moved `NodeCursorTrap` and `Overlays` into userland via the `system_components` config registry in svedit 0.1.0.

The feedback from #68 was clear: keep the library's default components minimal and exemplary, and ship advanced cursor behavior in app-land. This plan takes that direction further. Rather than iterating on the original `position: absolute` approach, it proposes a fundamentally different architecture based on CSS anchor positioning that solves the remaining structural issues. It's designed to be an improved default and/or a documented recipe for svedit developers.

### 1. `NodeCursorTrap` relies on `position: relative` from a `div`-only scoped style

`Node.svelte` applies `position: relative` via a scoped style that only targets `div` elements. `NodeCursorTrap` uses `position: absolute` to position itself relative to the parent node. When a block component renders a non-div tag (e.g. `<blockquote>`, `<figure>`, `<section>`), the cursor trap has no positioned ancestor and renders in the wrong location.

This also conflicts with CSS anchor positioning, where `position: relative` on ancestors can [cause subtle issues](https://css-tip.com/anchor-issues/).

### 2. `NodeCursorTrap` mixes three concerns into one component

The current component handles:
- **Keyboard caret target** (`.svedit-selectable` — the DOM element the browser's selection API lands on)
- **Visual cursor** (`.node-cursor` — the blinking bar the user sees)
- **Click target** (the `position: absolute` overlay that captures pointer events)

This coupling means each node in a `node_array` renders its own conditional visual cursor, orientation-detection `$effect` with resize event, and click handler. For a document with N nodes, that's N instances of all this logic, even though at most one cursor is ever visible. The memory usage is significant.

### 3. Hardcoded pixel values per orientation

The visual cursor and click target dimensions are specified separately for vertical and horizontal layouts using magic numbers (`-6px`, `12px`, `4px`, etc.) and `data-orientation` attribute selectors. Changing the gap between nodes, or switching to a new layout, requires updating multiple coordinated values. This is brittle and unergonomic for developers.

### 4. Blink animation doesn't match native caret

The upstream blink uses a smooth sine-like fade (`opacity: 0 → 1 → 0` over 0.7s). Native text carets use a slower, more organic rhythm. They also start with full visibility and stay visible while the pointer is pressed.


src/lib/Svedit.svelte
```diff
@@ -1128,12 +1128,13 @@ ${fallback_html}`;
...
-					inline: 'nearest'
-				});
-			}, 0);
+			// setTimeout(() => {
+			// 	(is_backward ? focus_node : anchor_node).scrollIntoView({
+			// 		block: 'nearest',
+			// 		inline: 'nearest'
+			// 	});
+			// }, 0);
+			// DISABLED THIS: IT FEELS QUITE SURPRISING TO HAVE THE SELECTION SCROLLED INTO VIEW WHEN THE USER CLICKS ON A NODE.
```
michael: This is needed because when you undo, you want to have the selection to be restored at where you were at that previous point in time.

johannesmutter:
So performance significantly improves with culling. Without the editor becomes unusable with 2000+ nodes. At 1000 nodes without culling the framerate is only ~7fps. With culling framerate is >40ps. Memory usage is also ca. 50% reduced: 117MB vs. 60MB. Render time is ¼  1100ms vs. 4000ms. With culling nodes count doesn't really matter, can easily render north of 10.000 nodes, though of course this depends on the types of nodes, but plaintext or e.g. svgs are no issue. 

https://github.com/user-attachments/assets/d3ea3358-a6bc-43a5-89fb-f5d85dba8bda


| Metric | OFF | ON (before cursor trap opt) | ON (with cursor trap opt) |
|--------|-----|----------------------------|---------------------------|
| DOM elements | 15,412 | 13,675 | **8,476** |
| Gap markers | 1,748 | 11 | 14 |
| Render (ms) | 3,389 | ~1,560 | **1,303** |
| FPS | 11 | ~28 | **60** |

Conditional cursor trap rendering removed **5,199 DOM elements** at 1000 nodes — that's a **38% reduction** from the previous ON state and a **45% reduction** from OFF. The render time dropped from ~1560ms to 1303ms, and FPS jumped to 60.

johannesmutter:
ok wait so the issue is then the currently active node-gap should not be removed in the culling optimisation?

johannesmutter:
> gaps are not computed for page.nav and page.footer (which are references of type: 'node')

This is fixed now. see https://github.com/michael/svedit/pull/209/changes/611001b5581763e9dfa2220c8e8ee3cb18a966c1

johannesmutter:
## Summary: NodeGap culling fix and performance investigation

### The fix

NodeGap elements are now **always present in the DOM** when editable, but anchor positioning is **lazily activated** via a `.positioned` class toggle driven by the existing IntersectionObserver/culler. Off-screen gaps get `position: static; width: 0; height: 0; overflow: hidden` — present in the DOM tree but with zero layout cost. When the culler detects a node near the viewport, it flips `positioned` to true, which activates the full CSS anchor positioning rules.

This fixes the selection bug: the browser's Selection API keeps a reference to the anchor node (the `.svedit-selectable` inside the gap where a drag started). Since gap elements are never removed from the DOM, that reference stays valid regardless of scrolling.

It also fixes the scrollTo issue from #214 — gap elements can always be targeted programmatically.

The DOM structure is now stable: `ui = f(doc_state, editable_state)` holds. No viewport-driven DOM insertions or removals.

### Benchmark results (7 configurations tested)

The recommended configuration is **Tests 1-2** (always in DOM, `.positioned` toggle, no `contain: strict`).

**Key comparison at 2000 nodes (Tests 1-2 vs Tests 3-4):**

| Metric | Always in DOM | Culled from DOM |
|---|---|---|
| DOM elements | 27,378 | 16,950 |
| Scroll avg FPS | 59 | 60 |
| Scroll min FPS | 30 | 57 |
| Resize avg FPS | 8-9 | 12 |
| Render ms | 1923-2001 | 1780-1790 |

Up to 1000 nodes, scroll performance is identical (60 FPS). The regression appears at 2000+ nodes, primarily as a scroll min-FPS spike (30 vs 57) and ~30% worse resize FPS. The scroll *average* stays at 59 FPS — the min-FPS drop is a single-frame spike during IO-driven class transitions.

### Optimization routes explored (and why they didn't work)

**1. `contain: strict` on un-positioned gaps (Tests 5-6)** — Catastrophic. Render time at 6000 nodes went from ~8.8s to ~48s (5-6x slower). The browser freezes for seconds at 2000 nodes. Likely cause: toggling between `contain: strict` and no containment triggers heavy style/layout recalculation, and containment boundaries may conflict with CSS anchor positioning resolution.

**2. Removing the `.positioned` toggle entirely (Test 7)** — Let the browser handle everything via `position-visibility: anchors-visible`. Much worse: resize became unusable at just 200 nodes (26 FPS), scroll dropped to 39 FPS at 500 nodes. **Key finding: `position-visibility: anchors-visible` does NOT skip anchor resolution.** The browser resolves all `anchor()` functions on every layout pass and then hides the result. The `.positioned` class toggle is essential — it's the only thing preventing O(N) anchor resolution.

**3. Conditional inline styles (only set `anchor-name` and CSS variables when positioned)** — Rejected after analysis. The `gap_style` and `anchor_name` derivations currently depend only on `path` (which never changes), so they compute once at mount and never re-evaluate. Making them conditional on `positioned` would *add* a reactive dependency that triggers re-evaluation on every scroll-driven visibility change, plus extra DOM mutations (style attribute updates) on top of the class toggle. Strictly worse during scroll.

**4. Short-circuiting `$derived` computations when not positioned** — Same problem as #3. Adds `positioned` as a reactive dependency to derivations that currently have none during scroll. The "optimization" would increase per-transition cost.

### Where the remaining performance gap comes from

The cost has two sources that can't be further optimized at this layer:

1. **~10,000 extra DOM elements at 2000 nodes** (~62% more). Even as zero-size static elements, the browser must include them in style recalculation passes. This primarily affects resize performance.

2. **~4,000 extra Svelte component instances** in the reactive graph. Each NodeGap has 2 `$derived` values. When the culler's version counter bumps during scroll, Svelte must check all Node.svelte instances' `is_near_viewport` derivations and diff the `positioned` prop on all NodeGap components. This is the main source of the scroll min-FPS spike at 2000 nodes.

### Ideas for further improvement

- **Reduce Svelte component overhead**: The NodeGap component could potentially be replaced with a Svelte action (`use:node_gap`) that creates gap DOM elements imperatively, bypassing Svelte's reactive graph entirely. This would eliminate ~8,000 reactive derivations from the graph. Trade-off: loses the `system_components.NodeGap` customization hook and is harder to maintain.

- **Fix foundational Svelte reactivity**: Michael's hypothesis about `$state.raw` causing O(N) re-renders on typing is worth investigating independently. If Svelte's reactivity can be made more local (only re-render changed nodes), the extra NodeGap instances become less impactful.

- **Accept the trade-off for now**: For web-page-scale documents (Wordpress-like), 200-500 nodes is typical. At those counts, performance is identical to the culled-from-DOM approach. The regression only appears at 2000+ nodes, where resize was already poor (12 FPS culled vs 9 FPS always-in-DOM).

<details>
<summary>Full Benchmark Results</summary>

### Test 1: NodeGaps always in DOM

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 79 | 673 | 21 | 28 | 60 | 56 | 60 | 57 |
| 100 | 99 | 1355 | 21 | 18 | 60 | 57 | 60 | 57 |
| 200 | 188 | 2718 | 21 | 22 | 60 | 56 | 60 | 57 |
| 500 | 571 | 6828 | 21 | 37 | 60 | 56 | 35 | 20 |
| 1000 | 1111 | 13685 | 21 | 67 | 60 | 56 | 18 | 12 |
| 2000 | 2001 | 27378 | 21 | 160 | 59 | 30 | 8 | 4 |
| 6000 | 8667 | 82172 | 15 | 333 | 34 | 15 | 2 | 1 |

### Test 2: NodeGaps always in DOM

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 76 | 673 | 21 | 29 | 60 | 57 | 60 | 57 |
| 100 | 91 | 1355 | 21 | 26 | 60 | 56 | 60 | 56 |
| 200 | 178 | 2718 | 21 | 22 | 60 | 54 | 60 | 56 |
| 500 | 473 | 6828 | 21 | 37 | 60 | 55 | 39 | 20 |
| 1000 | 909 | 13685 | 21 | 67 | 60 | 56 | 19 | 10 |
| 2000 | 1923 | 27378 | 21 | 166 | 59 | 30 | 9 | 5 |
| 6000 | 8939 | 82172 | 15 | 333 | 37 | 15 | 2 | 1 |

### Test 3: NodeGaps not in DOM when culled

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 79 | 484 | 21 | 14 | 60 | 56 | 60 | 56 |
| 100 | 93 | 899 | 21 | 16 | 60 | 56 | 60 | 57 |
| 200 | 166 | 1740 | 21 | 21 | 60 | 56 | 60 | 57 |
| 500 | 435 | 4275 | 21 | 38 | 60 | 56 | 49 | 20 |
| 1000 | 834 | 8504 | 21 | 58 | 60 | 56 | 26 | 15 |
| 2000 | 1790 | 16950 | 21 | 168 | 60 | 57 | 12 | 7 |
| 6000 | 8031 | 50726 | 15 | 323 | 47 | 15 | 3 | 1 |

### Test 4: NodeGaps not in DOM when culled

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 73 | 484 | 21 | 37 | 60 | 57 | 60 | 56 |
| 100 | 88 | 899 | 21 | 17 | 60 | 47 | 60 | 57 |
| 200 | 160 | 1740 | 21 | 21 | 60 | 57 | 60 | 58 |
| 500 | 435 | 4275 | 21 | 38 | 60 | 57 | 51 | 30 |
| 1000 | 826 | 8504 | 21 | 60 | 60 | 56 | 26 | 15 |
| 2000 | 1780 | 16950 | 21 | 155 | 60 | 57 | 12 | 6 |
| 6000 | 7840 | 50726 | 15 | 323 | 47 | 15 | 3 | 1 |

### Test 5: NodeGaps always in DOM, but `contain: strict` set

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 70 | 670 | 18 | 40 | 60 | 56 | 60 | 57 |
| 100 | 104 | 1355 | 21 | 17 | 60 | 57 | 60 | 57 |
| 200 | 193 | 2718 | 21 | 22 | 59 | 30 | 60 | 57 |
| 500 | 547 | 6828 | 21 | 39 | 60 | 56 | 39 | 20 |
| 1000 | 1255 | 13685 | 21 | 67 | 60 | 57 | 20 | 9 |
| 2000 | 3389 | 27378 | 21 | 159 | 59 | 30 | 9 | 4 |
| 6000 | 45405 | 82172 | 15 | 310 | 24 | 15 | 2 | 1 |

### Test 6: NodeGaps always in DOM, but `contain: strict` set

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 82 | 673 | 21 | 14 | 60 | 56 | 60 | 57 |
| 100 | 113 | 1355 | 21 | 26 | 60 | 56 | 60 | 56 |
| 200 | 221 | 2718 | 21 | 23 | 59 | 30 | 60 | 56 |
| 500 | 542 | 6828 | 21 | 39 | 60 | 56 | 39 | 20 |
| 1000 | 1223 | 13685 | 21 | 67 | 60 | 56 | 20 | 12 |
| 2000 | 3404 | 27378 | 21 | 114 | 59 | 30 | 9 | 5 |
| 6000 | 50677 | 82172 | 15 | 310 | 32 | 15 | 2 | 1 |

### Test 7: No positioned class toggle. All NodeGap elements always in DOM

| Nodes | Render (ms) | DOM els | Gap markers | Memory (MB) | Scroll avg | Scroll min | Resize avg | Resize min |
|---|---|---|---|---|---|---|---|---|
| 50 | 93 | 673 | 21 | 24 | 60 | 57 | 60 | 56 |
| 100 | 137 | 1355 | 21 | 17 | 60 | 56 | 50 | 30 |
| 200 | 228 | 2718 | 21 | 23 | 56 | 29 | 26 | 15 |
| 500 | 485 | 6828 | 21 | 36 | 39 | 12 | 10 | 9 |
| 1000 | 994 | 13685 | 21 | 62 | 32 | 7 | 5 | 3 |
| 2000 | 2451 | 27378 | 21 | 117 | 35 | 20 | 2 | 1 |
| 6000 | 12158 | 82172 | 15 | 318 | 9 | 7 | 0 | 0 |

</details>

PR #210, opened 2026-03-02
Centralize DOM path serialization helpers

johannesmutter:
## Summary
- add core helpers `serialize_dom_path(path)` and `deserialize_dom_path(path_key, context?)` in `src/lib/dom_path.js` to provide lossless, typed DOM path encoding
- migrate core `data-path` producers and consumers to use centralized helpers instead of ad-hoc `join('.')` / `split('.')`
- export the helpers via `src/lib/index.js` and add unit coverage in `src/test/dom_path.test.js` for roundtrip, numeric-string IDs, and legacy fallback behavior

## Test plan
- [ ] Run `npm run test:unit`
- [ ] In the demo, verify text, node, and property selections still map correctly after clicks/keyboard navigation
- [ ] In the demo, verify collapsed and expanded node selections still render and restore correctly

### DIFF
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -1,5 +1,6 @@
 <script>
 	import { getContext } from 'svelte';
+	import { serialize_dom_path } from './dom_path.js';
 	import { char_slice, get_char_length, snake_to_pascal, get_selection_range } from './utils.js';
 
 	/** @import { AnnotatedTextPropertyProps, Annotation, Fragment, SelectionRange } from './types.d.ts'; */
@@ -106,7 +107,7 @@
 <svelte:element
 	this={tag}
 	data-type="text"
-	data-path={path.join('.')}
+	data-path={serialize_dom_path(path)}
 	style="anchor-name: --{path.join('-')};{style}"
 	class="text svedit-selectable {css_class}"
 	class:empty={is_empty}
--- a/src/lib/CustomProperty.svelte
+++ b/src/lib/CustomProperty.svelte
@@ -1,5 +1,6 @@
 <script>
 	import { getContext } from 'svelte';
+	import { serialize_dom_path } from './dom_path.js';
 
 	/** @import { CustomPropertyProps } from './types.d.ts'; */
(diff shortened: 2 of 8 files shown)


johannesmutter:
`data-path` is the **address label** that connects DOM elements to your document model.  
If that label is encoded/decoded inconsistently, the editor can target the wrong place.

Before this PR, different places used ad-hoc `path.join('.')` and `split('.')`.  
That looks simple, but it creates ambiguity and hidden bugs.

---

### What can go wrong without centralized helpers

1. **Numeric-looking IDs get misread as indexes**
   - Example path: `["123", "body", 0, "items"]`
   - Naive parsing can turn this into: `[123, "body", 0, "items"]`
   - `"123"` (node ID string) is not the same as `123` (number index)
   - Result: failed lookups, missing overlays, wrong selection mapping

2. **Dots inside IDs break path boundaries**
   - Example ID: `"story.v2"`
   - Serialized with `join('.')`: `"story.v2.body.0"`
   - `split('.')` becomes: `["story", "v2", "body", "0"]` (wrong segments)
   - Result: editor points to wrong node/property or crashes on invalid path

3. **Different files parse paths differently**
   - Some places treat all segments as strings
   - Some parse certain segments as numbers
   - Some compare paths by string form
   - Result: subtle “works here, breaks there” behavior (selection, cursor traps, rendering)

---

### Why centralized helpers fix this

With `serialize_dom_path(path)` and `deserialize_dom_path(path_key, context?)`:

- path encoding is **typed and lossless**
- decoding logic is **single-source-of-truth**
- behavior is **consistent** across `Node`, `NodeArrayProperty`, `CustomProperty`, `AnnotatedTextProperty`, and `Svedit`
- future changes are safer (one codec to update, not many scattered `split/join` call sites)

---

### Practical impact

This prevents hard-to-debug bugs I encountered while improving the Overlays like:

- clicking a node selects nothing / wrong node
- node cursor appears in wrong place
- property/text selection fails to restore correctly
- overlays disappear unexpectedly for certain IDs

In short: the goal of this PR is to path handling reliable long-term.

michael:
Done via https://github.com/michael/svedit/pull/253

PR #212, opened 2026-03-06
Type-to-create, fix copy on collapsed node selection, clean up logging

johannesmutter:
## Summary
- **Type-to-create**: Typing at a collapsed node cursor now inserts a default block and places the typed character into its first `annotated_text` property, enabling a natural "just start typing" flow.
- **Fix copy on collapsed node selection**: The copy handler now early-returns for collapsed node selections (where there is nothing to copy), preventing errors.
- **Clean up logging**: All debug `console.log` calls are now gated behind `if (dev)` to avoid performance overhead in production. Leftover unconditional `console.log` statements in the paste handler have been removed.

## Test plan
- [ ] Place a node cursor (collapsed) and start typing — verify a new default block is created and text appears
- [ ] Select a collapsed node cursor and press Cmd+C — verify no error is thrown
- [ ] Verify no `console.log` output appears in production builds
- [ ] Copy/paste text and nodes still work as expected

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -1,4 +1,5 @@
 <script>
+	import { dev } from '$app/environment';
 	import { getContext, setContext } from 'svelte';
 	import {
 		snake_to_pascal,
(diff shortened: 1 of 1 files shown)


johannesmutter:
Quick note: I had these in the other Overlays PR and move them here for separate review.

PR #231, opened 2026-04-01
Virtual keyboard custom property

johannesmutter:
**Problem**

1. **iOS virtual keyboard opens when tapping non-text custom properties** (images, embeds)
2. **Scroll jumps to wrong position** when selecting a custom property

**Solution**

**Keyboard suppression:**

- `handle_pointerdown` sets `inputmode="none"` on the canvas when a touch lands on a property element, before the browser's focus logic runs
- `handle_canvas_click` sets the property selection on confirmed taps (click doesn't fire for scrolls)
- `suppress_focus` flag blocks `onselectionchange` and `handle_canvas_focus` from interfering during the tap event chain
- After a completed property tap, subsequent pointerdowns early-return to avoid DOM queries that re-trigger the keyboard on iOS during scroll
- Drag detection (`register_drag_detection`) unblocks `onselectionchange` when movement exceeds 5px, so drag-select across nodes still works
- State is cleaned up on blur, text taps, and new property taps

**Scroll fix:**

- `.property-selectable` uses `display: flex; align-items: flex-end` to position `.svedit-selectable` at the bottom of the property overlay — the browser's native scroll-to-caret now shows the full property instead of just its top edge
- `__render_property_selection` scrolls the property element itself rather than the DOM selection's focus node

### DIFF
--- a/src/lib/CustomProperty.svelte
+++ b/src/lib/CustomProperty.svelte
@@ -37,5 +37,11 @@
 		bottom: 0;
 		z-index: 1;
 		outline: none;
+		/* Position the hidden selectable element at the bottom so the
+		   browser's native scroll-to-caret ensures the full property
+		   is visible, not just the top edge. */
+		display: flex;
+		align-items: flex-end;
+		justify-content: center;
 	}
 </style>
--- a/src/lib/CustomProperty.svelte
+++ b/src/lib/CustomProperty.svelte
@@ -44,4 +44,7 @@
 		align-items: flex-end;
 		justify-content: center;
 	}
+	.svedit-selectable {
+		caret-color: transparent;
+	}
 </style>
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -40,6 +40,8 @@
 	let is_composing = $state(false);
 	let canvas_focused = $state(false);
 	let before_composition_selection = null;
+	let suppress_focus = false;
+	let pending_property_path = null;
 
 
 	// let is_mobile = $derived(is_mobile_browser());
@@ -262,6 +264,12 @@
 		if (!editable) return;
 		if (!canvas_focused) return;
 		if (is_composing) return;
+		// When a property tap is pending, ignore browser-initiated selection
+		// changes. The selection will be set by handle_property_click instead.
+		// Without this guard, the browser's focus/tap places a DOM selection
+		// that gets mapped to a text/node selection, triggering render_selection
+		// with scrollIntoView before the property click handler runs.
+		if (suppress_focus) return;
 		const dom_selection = window.getSelection();
 		if (!dom_selection.rangeCount) return;
 
@@ -749,8 +757,94 @@ ${fallback_html}`;
 		}
 	}
 
+	/** Resolve the property element at the given coordinates, or null. */
+	function property_el_from_point(x, y) {
+		if (!document.caretPositionFromPoint) return null;
+		const caret_pos = document.caretPositionFromPoint(x, y);
+		if (!caret_pos) return null;
+		const target = caret_pos.offsetNode instanceof HTMLElement
+			? caret_pos.offsetNode
+			: caret_pos.offsetNode?.parentElement;
+		return target?.closest('[data-path][data-type="property"]') ?? null;
+	}
+
+	function reset_suppress_focus() {
+		canvas_el.removeAttribute('inputmode');
+		suppress_focus = false;
+		pending_property_path = null;
+	}
+
+	/**
+	 * On pointerdown, suppress the virtual keyboard on iOS if the pointer
+	 * is over a property element. The actual selection is deferred to
+	 * handle_canvas_click so scroll gestures don't change the selection.
+	 */
+	function handle_pointerdown(event) {
+		if (!editable) return;
+
+		const property_el = property_el_from_point(event.clientX, event.clientY);
+
+		if (!property_el) {
+			if (suppress_focus) reset_suppress_focus();
+			return;
+		}
+
+		canvas_el.setAttribute('inputmode', 'none');
+		suppress_focus = true;
+		pending_property_path = property_el.dataset.path.split('.');
+
+		// If the user drags instead of tapping, reset suppression so
+		// onselectionchange can process the drag selection normally.
+		const start_x = event.clientX;
+		const start_y = event.clientY;
+		const DRAG_THRESHOLD = 5;
+		function on_pointermove(e) {
+			const dx = e.clientX - start_x;
+			const dy = e.clientY - start_y;
+			if (dx * dx + dy * dy > DRAG_THRESHOLD * DRAG_THRESHOLD) {
+				reset_suppress_focus();
+				canvas_el.removeEventListener('pointermove', on_pointermove);
+			}
+		}
+		canvas_el.addEventListener('pointermove', on_pointermove);
+		// Clean up the listener when the pointer interaction ends.
+		canvas_el.addEventListener('pointerup', () => {
+			canvas_el.removeEventListener('pointermove', on_pointermove);
+		}, { once: true });
+	}
+
+	/**
+	 * On click (only fires for completed taps, not scrolls), set the
+	 * property selection using the path cached from pointerdown.
+	 */
+	function handle_canvas_click() {
+		if (!editable || !suppress_focus || !pending_property_path) return;
+
+		const path = pending_property_path;
+		pending_property_path = null;
+
+		// Set both inside flushSync so the $effect sees them atomically.
+		// Otherwise canvas_focused=true triggers render_selection with
+		// the OLD selection (e.g. a text selection), causing scrollIntoView
+		// to scroll to the wrong place.
+		flushSync(() => {
+			canvas_focused = true;
+			session.selection = { type: 'property', path };
+		});
+	}
+
 	// Handle focus - push session's keymap onto stack
 	function handle_canvas_focus() {
+		// When a property tap is pending, skip the normal focus flow
+		// (which clears the selection and can cause unwanted scrolling).
+		// handle_canvas_click will set the selection instead.
+		if (suppress_focus) {
+			flushSync(() => {
+				canvas_focused = true;
+			});
+			key_mapper?.push_scope(session.keymap);
+			return;
+		}
 		// Use flushSync so highlight spans are removed from the DOM
 		// immediately, before the browser processes the click's selection.
 		flushSync(() => {
@@ -1187,6 +1281,7 @@ ${fallback_html}`;
 		const el = canvas_el.querySelector(
 			`[data-path="${selection.path.join('.')}"][data-type="property"]`
 		);
+
 		const gap_selectable = el.querySelector('.svedit-selectable');
 		const range = window.document.createRange();
 		const dom_selection = window.getSelection();
@@ -1199,10 +1294,7 @@ ${fallback_html}`;
 
 		// Scroll the selection into view
 		setTimeout(() => {
-			const selectedElement = dom_selection.focusNode.parentElement;
-			if (selectedElement) {
-				selectedElement.scrollIntoView({ block: 'nearest', inline: 'nearest' });
-			}
+			el.scrollIntoView({ block: 'nearest', inline: 'nearest' });
 		}, 0);
 	}
 
@@ -1364,6 +1456,8 @@ ${fallback_html}`;
 		{onbeforeinput}
 		{oncompositionstart}
 		{oncompositionend}
+		onpointerdown={handle_pointerdown}
+		onclick={handle_canvas_click}
 		onfocus={handle_canvas_focus}
 		onblur={handle_canvas_blur}
 		contenteditable={editable ? 'true' : 'false'}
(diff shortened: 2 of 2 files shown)


johannesmutter:
Tested now with my app and editable-website repo. no issues. Ready for merge

michael:
While doing some more testing here I randomly ended up in situations where I have the cursor inside some text but they keyboard didn't come back.

Took me a while so i can reproduce it:

- select a custom property
- scroll the page a bit
- select some text
- (keyboard doesn't come back; in order to get the keyboard back you have to select a custom proerty again - then not scroll - but click at some text)

> Drag detection (register_drag_detection) unblocks onselectionchange when movement exceeds 5px, so drag-select across nodes still works

I don't yet understand why we need drag detection. Instead of juggeling additional UI state (which sooner or later gets out of sync), couldn't we just have an invariant derived from the internalSvedit model. Something like:

```svelte
<div
  class="svedit-canvas"
  contenteditable="true"
  inputmode={session.selection?.type === 'property' ? 'none' : 'text'}
>
  ...
</div>
```

michael:
Also: can you give me the steps to reproduce the original `Scroll jumps to wrong position when selecting a custom property` issue?

johannesmutter:
> Also: can you give me the steps to reproduce the original `Scroll jumps to wrong position when selecting a custom property` issue?

The original here was: tap an image on iOS → page scrolls so only the top part of the image is visible (or the image is partially under the keyboard).

so on svedit.dev just scroll down and tap the first image. it's mostly out of view and the toolbar is not visible. Compare this to the behaviour on this branch.

Video shows this branch with fix, and current state on main afterwards
https://github.com/user-attachments/assets/9e870292-537b-4936-ad4e-bd1b747f2db2

johannesmutter:
> * select a custom property
> * scroll the page a bit
> * select some text
> * (keyboard doesn't come back; in order to get the keyboard back you have to select a custom proerty again - then not scroll - but click at some text)
> 
> >

Fix in last commit

michael:
Thanks for the fix.

I opened another PR #241 only with a "show-no-keyboard-on-property-selections" fix. Trying to test/understand if we really need the additional state tracking.

Can be tested here: https://svedit-git-no-keyboard-for-p-101cdb-michael-aufreiters-projects.vercel.app/

I can confirm that when I click the first image there, it scrolls out of view. For all subsequent clicks on images no weird scroll happens. So it's just that initial scroll, that's a bit annoying (and should be fixed I agree)

Wondering why that scroll-out-of-view happens, because first i assumed it's because of shifting the content because the soft keyboard slides into place, but there seems to be another root issue, which I'd like to figure out and tackle (still investigating, glad for any hints you have).

michael:
Additional context: We might want to solve the scroll into view issue by using a custom function, that will respect edge cases like element is already partially visible, which direction are you navigating from etc. Did something like that to improve scrollintoview behavior in the pages drawer:

https://github.com/michael/editable-website/blob/main/src/routes/components/PagesDrawer.svelte#L45

Maybe something similar can be done in Svedit-land, or ideally we find a generic scrollintoview method that works smoothly everywhere.

michael:
Applied your CSS fixes to CustomProperty, and it seems to fix the scroll out of view issue.

Test here: https://svedit-ft4lu40ur-michael-aufreiters-projects.vercel.app

Anything else that this PR covers, and #241 does not?

EDIT: some previous comment that had context was lost somehow. It stated that I figured out that scrollintoview was never triggered on the first click at an image, so the scrolling is native behavior I believe triggered by onselectionchange

johannesmutter:
Looks good, tested a bunch of different selection + scroll scenarios, especially the drag-select-from-property-into-text case. Works all fine without the custom suppress functions.

Added comments to document and clarify what the change on inputmode does. 
The CSS had already good documentation.

johannesmutter:
@michael  Do you want to test against Android as well?

johannesmutter:
Ok I just tried on an Android Tablet, and it works fine there too (Chrome Browser).

It's funny I used my simplify skill with Claude Code on the previous state of this branch and it landed on the same result.

...

Ready for merge?

johannesmutter:
> Additional context: We might want to solve the scroll into view issue by using a custom function, that will respect edge cases like element is already partially visible, which direction are you navigating from etc. Did something like that to improve scrollintoview behavior in the pages drawer:
> 
> https://github.com/michael/editable-website/blob/main/src/routes/components/PagesDrawer.svelte#L45
> 
> Maybe something similar can be done in Svedit-land, or ideally we find a generic scrollintoview method that works smoothly everywhere.

Agree that would be very handy. Ideally native/ just with css, so it's easily "cancelable" by the user.

michael:
Looks great! 

Before merging might just clarify some implications of this change:

Not having the soft keyboard shown always forces apps (like editable.website) to show tools for certain operations. E.g. you might still want to delete the selection (backspace key). So what I did in EW, is show that as a tool now.

image

The reason I use a "backspace" tool, rather than a "clear image" tool is that there are other situations on mobile where you want that too:

- when you closed the soft keyboard (so the canvas doesn't have focus anymore)
- for any node selection, where we probably want to hide the soft keyboard as well (no need to type anything here)

Consequently, apps might also want to add an "ENTER" (or "+") tool when you have a collapsed node selection (node caret).

Is that approach reasonable and favorable over the current "add nodes via ENTER and delete via BACKSPACE" approach?

If yes, this might affect your "type into a node cursor to create new blocks" approach (might be tolerable on mobile though).

Wdyt?

johannesmutter:
"type into a node cursor to create new blocks" still works, since the node-gap is a "textual" selection. Just tested on my interface lineage project.

I don't foresee any other issues. If something does come up, we could always make this customisable on the app layer rather than hardcoding the mapping. A config-shaped extension point following the existing `session.config` pattern could look like this:

```javascript
// app side
session.config.selection_types = {
    text:     { inputmode: 'text' },
    node:     { inputmode: 'text' },
    property: { inputmode: 'none' },
    // app can add custom keys the library doesn't know about, e.g.:
    // property: { inputmode: 'none', toolbar: ImageToolbar, allow_drag: false }
};
```

But right now I don't see the need for configuration need in app land.

---

**Aside on mobile UX:**

I think mobile needs fundamentally different affordances than the desktop toolbar, not just a smaller or rearranged version of the same controls, or moving buttons from the physical keyboard to a "digital one". It works as a quick fix, but just doesn't feel great and clutters the UI.

A good example is the [Craft app](https://www.craft.do/de), Things, or even Apple Notes: deleting a block is a long swipe from the right; selecting one or multiple blocks is triggered by a subtle short swipe that overlays checkboxes, which works much better on mobile than the drag-select we have today. It feels gesture-native rather than ported.

The negative counter-example is the Notion iOS app: half a dozen operations are simply unavailable on mobile because there's no counterpart for them, and the ones that are available are just taken 1:1 from desktop. The result feels cramped and frustrating.

Worth thinking about as a separate track once the editing primitives are fully stable.

johannesmutter:
The virtual keyboard sure solves backspace delete/ enter to create, but is has so many downsides
- 50 % of the screen just gone
- out of 46 keys my keyboard shows, only 2 are relevant to the available actions 
- custom toolbar disappear below the virtual keyboard, and unless you build a native app wrapper, there's nothing we can do about it.

michael:
> "type into a node cursor to create new blocks" still works, since the node-gap is a "textual" selection. Just tested on my interface lineage project.

in the current version yes, but I thought we might want to hide the keyboard for all node selections too, which means we would only show it for text selections:

```html
<div
  class="svedit-canvas"
  inputmode={session.selection?.type === 'text' ? 'text' : 'none'}
>
  ...
</div>
```


> I think mobile needs fundamentally different affordances than the desktop toolbar, not just a smaller or rearranged version of the same controls, or moving buttons from the physical keyboard to a "digital one". It works as a quick fix, but just doesn't feel great and clutters the UI.
> 
> A good example is the [Craft app](https://www.craft.do/de), Things, or even Apple Notes: deleting a block is a long swipe from the right; selecting one or multiple blocks is triggered by a subtle short swipe that overlays checkboxes, which works much better on mobile than the drag-select we have today. It feels gesture-native rather than ported.
> 
> The negative counter-example is the Notion iOS app: half a dozen operations are simply unavailable on mobile because there's no counterpart for them, and the ones that are available are just taken 1:1 from desktop. The result feels cramped and frustrating.
> 
> Worth thinking about as a separate track once the editing primitives are fully stable.

Yeah I think (for now) I'm okay that EW is primarily aimed at Desktop editing, but I'm actually quite positively surprised that with Svedit/EW we are getting quite close to a solid mobile experience. Not perfect for sure and not comparable with a mobile-first design like Craft. So maybe to reframe my question:

What do you think could be done within a day or two of effort to optimize the mobile experience based on what we have right now? 

Tools could be re-arranged/re-labelled but under the hood they should drive the same ops we have on the desktop.

Best thing I could come up with so far is to expose `⌫` (delete/clear) and `+` (insert/replace) as a tool (maybe generally in addition to the keybindings on desktop too, as I saw people struggling with what to do with a blinking node cursor, or a selected image. If going with this route, my main concern atm would be the toolbar becoming too crowded. I do also see some positive sides in replicating the UX from desktop, as you'd basically know how everything works/find every tool you are familiar with from desktop usage, as opposed to discovering/learning different ux patterns for mobile (swipes and drags etc.).

EDIT below (didn't read your follow-up comment)

> The virtual keyboard sure solves backspace delete/ enter to create, but is has so many downsides
> 
> * 50 % of the screen just gone
> * out of 46 keys my keyboard shows, only 2 are relevant to the available actions
> * custom toolbar disappear below the virtual keyboard, and unless you build a native app wrapper, there's nothing we can do about it.

Fully agree with that. And that's what I'm proposing: Show the soft keyboard less often (=only when cursor is inside a text, and you are expected to type characters).

EDIT 2: 

Once soft keyboard is collapsed for say multi-node selections, we can expose a couple of ops that make sense in that scenario: 

- cut
- copy
- delete
- insert (overwrite selection with a new empty block) - not sure about this one

Assuming you have a svedit node-selection in the clipboard and clicking a node cursor we can then offer:
- paste (what's in the clipboard)
- delete (will delete the block before the cursor)
- insert (will create a new block at the cursor position)


Would make the copy&paste workflow much more smooth (and discoverable) than it is now, with the fiddly native cut/paste menu.

johannesmutter:
> What do you think could be done within a day or two of effort to optimize the mobile experience based on what we have right now?

Honestly: I’d defer rather than ship a half-fix. Or focus one gesture only: swipe-to-delete on a selected node. That's probably doable in a day.

Mobile users don't compare mobile Svedit to desktop Svedit. They compare it to every other mobile app they've used/ they are using in that moment/ they are changing from/to in parallel. The mental state when you use an app is shaped by all the other apps open, not by some pattern you used on a desktop "category equivalent tool" two days ago. Swipes, long-press, drag handles aren't things they need to “learn”; they're already the expectation. Showing a + and ⌫ in a toolbar is familiar from desktop Svedit/ a physical keyboard, but unfamiliar in the mobile context they're actually arriving from.

There's also a path-dependence cost: once you/ I ship those toolbar buttons, we've taught users where to find those operations. Adding swipe-to-delete later then becomes a change of behavior instead of the original gesture, which is much harder to roll out cleanly.

---

Otherwise: marking mobile editing as "not yet supported" in the README costs nothing and frees you to do it properly when there's time.

---

> under the hood they should drive the same ops we have on the desktop.

💯

michael:
I think both paths are worth exploring. I feel rather confident about replicating the desktop UX to mobile, and I think with this PR Svedit is ready for a "desktop-first mobile experience". It's maybe a matter of preference too: Might prefer select + contextual tool click to in-canvas swipe gesture, as much as i prefer cut+paste over drag and drop for reordering.

I could imagine that for EW this approach would stick for a long time. BUT: I might also be totally wrong, and I'll see users struggle.

Only one way to find out: We should try both.

I'd suggest we merge the current state of the PR, and I work on a new one that disables the soft keyboard on node selections as well.

Then I'll explore the "desktop-first mobile experience" in EW, while you try the other direction with more typical mobile gestures. My feel is this could all live in app-land and we can keep Svedit agnostic (might be wrong, and then we can add missing pieces to Svedit). No rush on this from your side. Feel free to defer mobile in your project and pick it up again whenever the time is right.

michael:
Regarding „swipe to delete“ - i was concerned this gesture would conflict with changing the native selection but seems you can natively select and then swipe left/right without destroying/changing the selection. So yeah, i guess swipe to delete might not be a biggie to implement.

also: why not have both, swipe to delete and an explicit delete tool? The former is faster the latter more discoverable?

johannesmutter:
> also: why not have both, swipe to delete and an explicit delete tool? The former is faster the latter more discoverable?

yes definitively, but space constraints often force a choice. Other actions may have higher priority for the limited UI real estate. iOS Mail is a good example: swipe gestures (delete, move, etc.) work everywhere, including list view. But the explicit Delete button only appears in full message view, and Move is tucked into a nested menu.

One can always cram everything into a scrollable toolbar — Notion does this on iOS — but then one sacrifices speed and simplicity. The tool becomes a power tool, which intimidates new users.

It follows a similar philosophy to: "start with keyboard shortcuts for everything, then introduce visible toolbar equivalents only where they earn their place."

here it would be:
"start with gestures for everything, then introduce visible toolbar equivalents only where they earn their place."

johannesmutter:
> I'd suggest we merge the current state of the PR, and I work on a new one that disables the soft keyboard on node selections as well.

Sounds good to me 👍

PR #232, opened 2026-04-01
Fix: Guard against the browser briefly reporting an out-of-bounds offset

johannesmutter:
Experienced this error within an app where I had type to create inside a node gap implemented, and when pressing the character `^` would trigger an error

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -273,7 +273,15 @@
 			// structurally identical — prevents a redundant $effect cycle
 			// (render_selection → scrollIntoView) on every DOM layout change.
 			if (JSON.stringify(selection) === JSON.stringify(session.selection)) return;
-			session.selection = selection;
+			// Guard against the browser briefly reporting an out-of-bounds offset
+			// during dead-key / IME composition (e.g. ^ key on European keyboards).
+			// The selection will be corrected on the next selectionchange event once
+			// the DOM has settled, so silently dropping the invalid one is safe.
+			try {
+				session.selection = selection;
+			} catch {
+				// ignore transient out-of-bounds selection
+			}
 		}
 	}
 


michael:
FYI: I activated the _validate_selection check for the session.selection so i capture missing missing edge cases.

So just eating the error will silence this. Does the error when it happens break the editor? If it's just printing the error on console and the editor continues working, then this this is just doing its job.

Does the error happen when the app is linked against the latest npm release, or only when linked against your current PR?

johannesmutter:
>  Does the error when it happens break the editor?

~It rejects the input and an error is logged in console:
e.g. `Uncaught Error: Text selection anchor_offset (1) is out of bounds. Max is 0.`~
Editing is still possible afterwards.

CORRECTION:
- for inside the AnnotatedText it just logs the error, no issue with editing afterwards and the character will be inserted, e.g. I type `^`, then `a` and it will add `â` 
- for inside NodeGap: the event is ignored and error is logged


> Does the error happen when the app is linked against the latest npm release, or only when linked against your current PR?

Happens with the latest state on main branch (via alias)

michael:
Sorry just to clarify, are we talking about editable-website, or another app? Can I reproduce it somehow. Would like to fix the root cause of this.

johannesmutter:
happening in my app, but you can also reproduce in editable website at position zero of a text node

michael:
Can reproduce, thanks. Yeah that's a tricky thing here, because technically we are in a wrong selection state. I think the correct fix is to not update the model selection while is_composing is true.

michael:
Fixed via https://github.com/michael/svedit/commit/38483fad175842ef8061052499fdcb71cc021d18


> for inside NodeGap: the event is ignored and error is logged

before the fix, couldn't reproduce the error with starting composition in a node gap. maybe you can check the version with the fix if you still see the problem.

michael:
Ah because you hav "type to create" there. Anyways, probably its fixed now.

johannesmutter:
Yes, can confirm fixed now!

PR #237, opened 2026-04-24
Node gap improvements

johannesmutter:
**This branch fixes a cluster of node-gap / gap-marker issues around wrapping layouts, scroll containers, and long documents.**

### What was broken

- **Wrong positioning in overflow-x / overflow-y containers** — gaps and markers used fixed-viewport anchor resolution that ignored the scroll offset of an ancestor scroll container.
- **Giant gap overlay** — when an anchor went orphan mid-edit, `anchor()` fell back to *Implicit Anchor Covers Invalid Target* → `top:auto` → the overlay expanded to cover the whole viewport.
- **Flicker and layout thrash on long documents** — every gap re-computed from scratch on every doc tick; no caching, no viewport culling.
- **Markers missing from DOM in wrap layouts** (flex-wrap, grid with `--row > 1`) — NodeArrayProperty rendered gaps for all positions but NodeGapMarkers rendered only a subset.
- **Gaps/markers remained visible after scrolling out of the node-array** even though they were no longer intersecting the array's edges.

### What changed

**`node_gap_computation.svelte.js` — rewritten around an IntersectionObserver:**
- Two-level gap cache keyed on `{ version, doc, node_ids }` per array_path, so unchanged arrays skip recomputation.
- Per-array version counter (`array_ver_map`) + per-path reactive signals distributed via `$effect.pre`.
- 500px overscan rootMargin, threshold `[0, 1]`, clip-bits skip.
- MutationObserver tracks add/remove, rAF-deferred.
- 20ms debounced version bumps to coalesce scroll storms.
- **Sync add helper** (`sync_add_if_visible`) runs a `getBoundingClientRect` viewport check at setup and on every MO addition, so nodes enter `index_map` in the same flush cycle as mount. Bypasses the 20ms debounce when sync-adds occur, so consumers never render from a stale cache.
- Removed a `past_overscan` DOM-order shortcut that was invalid in wrap layouts (DOM order ≠ visual top order).

**`NodeGap.svelte` / `NodeGapMarkers.svelte`:**
- Added `9999999px` fallback to all `anchor()` calls on `--_s-*` and `--_b-*` so orphan anchors collapse predictably instead of triggering IACVT.
- Bumped the branch-disable sentinel from `99999px` to `9999999px` to match fallback magnitude.
- Use `position-visibility: anchors-visible` so markers hide when their anchor scrolls out of the array's edges.

**`NodeArrayProperty.svelte`:**
- `{#each}` keyed by `node.id` (was index) to stabilize identity across reorders.
- Split `node_ids` / `nodes` derivations so array length reads don't churn.
- Empty-array gap now goes through `should_position_gap` like the rest.

### Scope
4 files, no API changes, no schema changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,6 +948,16 @@ Svedit provides specialized components for rendering different property types:
 />
 ```
 
+> **One path = one DOM mount.** Within a single Svedit document, each node path
+> may be mounted exactly once. Mounting the same path twice (e.g. rendering the
+> same `nav_items` array in both header and footer) breaks anchor positioning,
+> intersection tracking, id uniqueness, and selection mapping.
+> Svedit logs an error if it detects a duplicate mount.
+>
+> If you need to render shared content in multiple places, model it as distinct
+> node_arrays (`header_nav`, `footer_nav`) in the same document, or render it
+> from a separate Svedit instance.
+
 **`<CustomProperty>`** - For custom properties like images or other non-text content:
 
 ```svelte
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,22 +81,22 @@
 			"license": "MIT"
 		},
 		"node_modules/@emnapi/core": {
-			"version": "1.9.1",
-			"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.9.1.tgz",
-			"integrity": "sha512-mukuNALVsoix/w1BJwFzwXBN/dHeejQtuVzcDsfOEsdpCumXb/E9j8w11h5S54tT1xhifGfbbSm/ICrObRb3KA==",
+			"version": "1.10.0",
+			"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+			"integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
 			"dev": true,
 			"license": "MIT",
 			"optional": true,
 			"peer": true,
 			"dependencies": {
-				"@emnapi/wasi-threads": "1.2.0",
+				"@emnapi/wasi-threads": "1.2.1",
 				"tslib": "^2.4.0"
 			}
 		},
 		"node_modules/@emnapi/runtime": {
-			"version": "1.9.1",
-			"resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.9.1.tgz",
-			"integrity": "sha512-VYi5+ZVLhpgK4hQ0TAjiQiZ6ol0oe4mBx7mVv7IflsiEp0OWoVsp/+f9Vc1hOhE0TtkORVrI1GvzyreqpgWtkA==",
+			"version": "1.10.0",
+			"resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+			"integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
 			"dev": true,
 			"license": "MIT",
 			"optional": true,
@@ -106,13 +106,12 @@
 			}
 		},
 		"node_modules/@emnapi/wasi-threads": {
-			"version": "1.2.0",
-			"resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.0.tgz",
-			"integrity": "sha512-N10dEJNSsUx41Z6pZsXU8FjPjpBEplgH24sfkmITrBED1/U2Esum9F3lfLrMjKHHjmi557zQn7kR9R+XWXu5Rg==",
+			"version": "1.2.1",
+			"resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
+			"integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
 			"dev": true,
 			"license": "MIT",
 			"optional": true,
-			"peer": true,
 			"dependencies": {
 				"tslib": "^2.4.0"
 			}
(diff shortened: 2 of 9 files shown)


johannesmutter:
Note: in documents that are taller than 9999999px, the node gap markers and node gaps will not render correctly, as the formula then doesn't work anymore

michael:
> {#each} keyed by node.id (was index) to stabilize identity across reorders.

With this change Svedit would no longer be able to handle node arrays that reference the same node twice. We had that as a requirement originally to allow "real graphs" and "shared nodes", and the there's a unit test with that case that is now failing.

We could question that requirement, but it should be a intentional decision not an accidental one.

BTW: The possibility to have the same node occur in different places also led to the paths being index-driven, not id-driven.

_____

Again, if we changed our mind on this and say in Svedit within one document each node must occur/render only once, keying by node.id is okay. It would also make sure that anchored urls like #some_node_id will always point to one place.

Might make things easier for us now, but might as well be limiting for future use-cases.

Wdyt?

michael:
Tested with editable-website#main

- gaps inside overflowing containers (like the nav) are not immediately visible
- appear when scrolled out of view and scrolled back into view
- flicker while you are typing text in that area

https://github.com/user-attachments/assets/1f396939-ae52-4ca1-86b1-e14ebe839917

johannesmutter:
> > `{#each}` keyed by `node.id` (was `index`) to stabilize identity across reorders.
>
> With this change Svedit would no longer be able to handle node arrays that reference the same node twice. We had that as a requirement originally to allow "real graphs" and "shared nodes", and there's a unit test with that case that is now failing.

I think it's quite important that node arrays can reference the same node twice. Already using this pattern several times. So current state im this branch is obviously not acceptable.

### Two options

**A.** Revert to `{#each nodes as node, index (index)}`

**B.** Compound key with occurrence suffix:

```svelte
<script>
  let node_keys = $derived.by(() => {
    const counts = new Map();
    return node_ids.map(id => {
      const n = counts.get(id) ?? 0;
      counts.set(id, n + 1);
      return n === 0 ? id : `${id}#${n}`;
    });
  });
</script>

{#each nodes as node, index (node_keys[index])}
```

#### Tradeoff

| Scenario | A: `(index)` | B: compound |
|---|---|---|
| Duplicate nodes | ✅ | ✅ |
| Block identity | slot position | node instance |
| Mid-insert, mixed types | component swap at mismatched slots | block moves intact |
| Native DOM state on shifted slot | disturbed; rescued by `render_selection` | preserved |
| MO traffic on insert | minimal | more (detach + re-attach) |
| Code | simplest | +one `$derived.by` |

### Decisive question

Does any editing flow rely on the browser holding DOM state across a sibling insert (mid-composition typing, IME, caret across reorders)?

johannesmutter:
Ok seems like **No** is the answer to before question

...

- `onbeforeinput` early-returns while `event.isComposing` — IME writes natively; `oncompositionend` rolls it back and applies a controlled `insert_text()`.
- `insert_text()` only mutates the focused text node's `content`; siblings never shift.
- Sibling inserts (`break_text_node`, paste, programmatic) require explicit user actions that can't fire during composition.
- No programmatic-insert-while-typing pattern in the codebase.

Shifted slots never hold live DOM state at insert time.

...

I'll revert to A for now.

We can revisit C if a future feature inserts nodes while a text node is focused (e.g. collaborative remote ops mid-composition).

...

Let me know what you think

michael:
Can‘t yet really follow regarding the original problem here. Maybe you can give a full example why keying by index was problematic?

is this related to your type-at-nodecursor functionality or generally?

johannesmutter:
> Tested with editable-website#main
> 
> * gaps inside overflowing containers (like the nav) are not immediately visible
> * appear when scrolled out of view and scrolled back into view
> * flicker while you are typing text in that area
> 
>  gaps-flicker.mov

Your Nav renders the **same NodeArrayProperty twice** — once in the editable branch (overflow-x-auto), once in the public branch (hidden via `class:hidden`). Both are always in the DOM; only visibility toggles. 

### Quick Fix (in editable-website)

Render only one NodeArrayProperty at a time. Replace `class:hidden` with `{#if}` / `{:else}`:

```svelte
{#if show_edit_nav}
  <div class="{TW_LIMITER} overflow-x-auto overflow-y-hidden">
    …
    <NodeArrayProperty class="nav-items …" path={[...path, 'nav_items']} />
  </div>
{:else}
  <div class="{TW_LIMITER}">
    …
    <NodeArrayProperty class="nav-items …" path={[...path, 'nav_items']} />
  </div>
{/if}
```

Or factor the `<NodeArrayProperty>` into a snippet and render it once inside whichever wrapper is active.

...

but we should probably make svedit here more defensive. So that ids are unique across same instances and intersection observer and anchor positioning doesn't fail because there are duplicate ids or anchor names

michael:
Ah need to check, thought that the mobile branch isn‘t using NodeArray, as it‘s not meant to be editable.

but yeah generally people may render the same node array twice, which will lead to ambiguous node paths. We might want to just throw an error in that case.

johannesmutter:
The bigger issue here: **paths are the wrong identity**

A path is a data location (where in the document tree the node lives). A DOM element is a render context (where in the rendered UI it appears). Svedit currently conflates them: `data-path, id={node.id}, anchor-name: --{path}`, gap signal keyed by path, IO tracking keyed by path, querySelector by path, …

That conflation is fine when paths render exactly once, breaks when they don't. Header/footer nav with same nav data is a valid use case and counterexample. The same applies to any future "shared component" pattern (sidebar previews, split views, mini-maps).

johannesmutter:
hmm my "paths are wrong identity" framing is a bit overshooting …

You're right: within one document, paths *are* unique data locations by construction. The conflation only breaks when one path is mounted twice in the DOM. That concerns rendering only, not the data model.

So a contract could look like this: *one path = one DOM mount per document.* Throw a dev-mode error when a second NodeArrayProperty (or Node) registers an already-mounted path.

The use cases I was worried about have cleaner solutions at the data layer:

- **Header + footer nav with the same items** → distinct node_arrays (`header_nav`, `footer_nav`) in the same doc, or two Svedit documents. Either is more honest about intent than reusing one path twice.
- **Multi-document on one page** → genuinely needs a doc-id prefix on `data-path` / `anchor-name` / `id`. **We need to fix this separately!**

Wdyt?

johannesmutter:
> Multi-document on one page → genuinely needs a doc-id prefix on data-path / anchor-name / id. We need to fix this separately!

Actually a non-issue. doc id is already part of path. So 2x different documents on one page have different document_ids → no collision.

And if a developer ever does spawn two Sessions with the same document_id, they'd see the same symptoms as a duplicate-mount within one session. The existing duplicate-mount error in [#239](https://github.com/michael/svedit/issues/239) would catch that too. It'd fire on the second NodeArrayProperty mounting the same path_str. So the ergonomics are already covered.

michael:
Yes I also think that max one mounted component per path that exists in a document is a totally valid constraint/invariant we should enforce.

michael:
Anything left here, or good to merge now?

johannesmutter:
good to merge

PR #239, opened 2026-04-24
Enforce one-path-one-mount invariant for NodeArrayProperty

johannesmutter:
## Summary

Within a single Svedit document, each `NodeArrayProperty` path must be mounted in the DOM exactly once. Mounting the same path twice (e.g. rendering the same `nav_items` array in both header and footer, or toggling two copies with `class:hidden`) breaks anchor positioning, IntersectionObserver tracking, `id` uniqueness, gap signals and selection mapping.

This PR turns that assumption into an enforced invariant with a clear developer error, plus a cheap defensive CSS rule so the failure mode is less catastrophic if it slips through.

See discussion: michael/svedit#237 (comment sequence ending at [this comment](https://github.com/michael/svedit/pull/237#issuecomment-4316728543)).

## Changes

- **`Session.svelte.js`** — add `register_mount` / `unregister_mount` methods backed by a per-session `Set<string>`. In dev mode, the first registration wins and any subsequent mount logs an actionable `console.error` naming the path and the recommended alternatives. Gated on `esm-env`'s `DEV` constant (same flag Svelte itself uses internally).
- **`NodeArrayProperty.svelte`** — `$effect` that registers `path_str` on mount and unregisters on destroy. Also adds `anchor-scope: all` on `[data-type="node_array"]` as defensive isolation so per-node anchor-names can't cross-resolve across siblings (multi-document pages, or accidental duplicate mounts while the dev finds the root cause).
- **`README.md`** — documents the invariant and points to the two clean solutions for legitimate "same content, multiple places" use cases: distinct `node_array`s in the same doc, or separate Svedit instances.
- **`package.json`** — declares `esm-env` as a direct dependency (was transitive via Svelte).

## Why this instead of a path + render-context refactor?

Within one document, paths **are** unique data locations by construction. The conflation between "data location" and "render context" only breaks when a single path is mounted twice, and that is a render-time misuse with cleaner alternatives at the data layer. Throwing on duplicate mount preserves path-as-identity everywhere (`data-path`, `id`, `anchor-name`, IO tracking, gap signals, selection mapping) without introducing a render-context layer. Multi-document-on-one-page is a separate concern and would be handled with a doc-id prefix when we need it.

## Base branch

Targets `node-gap-improvements` (#237) because NodeArrayProperty's current `path_str` split-derivation only exists on that branch. GitHub will auto-retarget to `main` once #237 merges.

## Test plan

- [x] Existing preview (`npm run dev`) renders cleanly — no console errors, 10 node_arrays mount with `anchor-scope: all` applied.
- [x] Intentionally render the same `<NodeArrayProperty path={...} />` twice in a route; console shows the duplicate-mount error naming the path.
- [x] Production build (`npm run build`) — dev-mode check and `Set` tree-shake out; no `esm-env` import of `DEV` prevents DCE.
- [x] `package.json`: `esm-env` resolves and `npm install` is clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,6 +948,16 @@ Svedit provides specialized components for rendering different property types:
 />
 ```
 
+> **One path = one DOM mount.** Within a single Svedit document, each node path
+> may be mounted exactly once. Mounting the same path twice (e.g. rendering the
+> same `nav_items` array in both header and footer) breaks anchor positioning,
+> intersection tracking, id uniqueness, and selection mapping. In dev mode
+> Svedit logs an error if it detects a duplicate mount.
+>
+> If you need to render shared content in multiple places, model it as distinct
+> node_arrays (`header_nav`, `footer_nav`) in the same document, or render it
+> from a separate Svedit instance.
+
 **`<CustomProperty>`** - For custom properties like images or other non-text content:
 
 ```svelte
--- a/README.md
+++ b/README.md
@@ -951,7 +951,7 @@ Svedit provides specialized components for rendering different property types:
 > **One path = one DOM mount.** Within a single Svedit document, each node path
 > may be mounted exactly once. Mounting the same path twice (e.g. rendering the
 > same `nav_items` array in both header and footer) breaks anchor positioning,
-> intersection tracking, id uniqueness, and selection mapping. In dev mode
+> intersection tracking, id uniqueness, and selection mapping.
 > Svedit logs an error if it detects a duplicate mount.
 >
 > If you need to render shared content in multiple places, model it as distinct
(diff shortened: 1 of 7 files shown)


michael:
Added the same check for AnnotatedTextProperty and CustomProperty, as there you should also not re-mount paths.

This breaks selection mapping:

```css
  [data-type="node_array"] {
    anchor-scope: all;
  }
```

Don't think we want the defensive path here, better just generally fail with an error, also in production. Then we don't need to check for the DEV env. I think it's fine to have this book-keeping around at runtime too.

johannesmutter:
done!

PR #241, opened 2026-04-26
No keyboard for property selections

michael:
Aims to solve the same as https://github.com/michael/svedit/pull/231, without introducing additional state handling.

Let's compare and collect downsides of this solution over #231.


michael:
Closing in favor of https://github.com/michael/svedit/pull/231

PR #245, opened 2026-04-29
Cycle z-index on-overlapping node-gaps

johannesmutter:
Issue:
e.g. for nested node arrays within other node arrays with no padding/ margin, the node-gaps can overlap. To still allow clicks on the parent or child we can cycle through them on each click.

this change uses a stateless cycle_z_index function that queries the DOM for the previous [data-sent-to-back] element, removes the attribute, then sets it on the clicked element.

A CSS rule `.positioned :global(.svedit-selectable[data-sent-to-back])` lowers z-index to 0, keeping it above regular content but below sibling gaps at z-index: 1, enabling the change on the next click

### DIFF
--- a/src/lib/NodeGap.svelte
+++ b/src/lib/NodeGap.svelte
@@ -1,3 +1,15 @@
+<script module>
+	function cycle_z_index(/** @type {MouseEvent} */ e) {
+		const el = /** @type {HTMLElement} */ (e.currentTarget);
+		if (el.hasAttribute('data-sent-to-back')) {
+			document.querySelectorAll('.svedit-selectable[data-sent-to-back]').forEach(prev => {
+				prev.removeAttribute('data-sent-to-back');
+			});
+		}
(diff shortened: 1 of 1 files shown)


michael:
Seems reasonable, what's the quickest way to test this in Svedit (like provoke gaps to overlap)?

> this change uses a stateless cycle_z_index function that queries the DOM for the previous [data-sent-to-back] element, removes the attribute, then sets it on the clicked element.

Solution is fine I think, but the function is not stateless. It adds state to the DOM. Will this work with 3 or more gaps overlapping as well?

johannesmutter:
Yes, create 3x or more nested node arrays, then click the node gap.

With stateless I meant not svelte $state

johannesmutter:
Now works with multiple overlapping node gaps, not just binary toggle. In this example there are 4 overlapping node gaps:


https://github.com/user-attachments/assets/dd552d62-1eeb-4910-85d9-68c8cdb51b17

PR #250, opened 2026-04-29
AnnotatedTextProperty: text-box: trim; where() clauses; caret alignment fix

johannesmutter:
- reduce specificity of classes, so no conflicts with :global() rules.
- make trim-both cap text the default for texts on the web. 
- fix bug in chrome with misaligned caret in empty contenteditable when text-box is used

### DIFF
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -162,8 +162,8 @@
 
 	/* Fix Chrome bug: caret correct on baseline in empty contenteditable. */
 	:where(.text.editable.empty) {
-		line-height: 1cap;
-		min-height: 1cap;
+		line-height: 1cap !important;
+		min-height: 1cap !important;
 	}
 
 	/* We switch from ::before to ::after when the element is focused. So the the caret is always before the placeholder. */
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -146,11 +146,26 @@
 </svelte:element>
 
 <style>
-	.text {
+	/* Editable text base layout; :where() allows easy override without specificity conflicts. */
+	:where(.text) {
 		white-space: pre-wrap;
 		overflow-wrap: anywhere;
 		box-sizing: content-box;
 	}
+
+	/* Trim extra vertical space for precise padding; keeps descenders visible. Considered a new baseline for typography on the web. */
+	@supports (text-box: trim-both cap text) {
+		:where(.text) {
+			text-box: trim-both cap text;
+		}
+	}
+
+	/* Fix Chrome bug: caret correct on baseline in empty contenteditable. */
+	:where(.text.editable.empty) {
+		line-height: 1cap;
+		min-height: 1cap;
+	}
+
 	/* We switch from ::before to ::after when the element is focused. So the the caret is always before the placeholder. */
 	[placeholder].empty:not(.focused)::before,
 	[placeholder].empty.focused::after {
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -160,12 +160,6 @@
 		}
 	}
 
-	/* Fix Chrome bug: caret correct on baseline in empty contenteditable. */
-	:where(.text.editable.empty) {
-		line-height: 1cap !important;
-		min-height: 1cap !important;
-	}
-
 	/* We switch from ::before to ::after when the element is focused. So the the caret is always before the placeholder. */
 	[placeholder].empty:not(.focused)::before,
 	[placeholder].empty.focused::after {
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -153,10 +153,10 @@
 		box-sizing: content-box;
 	}
 
-	/* Trim extra vertical space for precise padding; keeps descenders visible. Considered a new baseline for typography on the web. */
-	@supports (text-box: trim-both cap text) {
+	/* Trim extra vertical space for precise padding; Considered a new baseline for typography on the web. */
+	@supports (text-box: trim-both cap alphabetic) {
 		:where(.text) {
-			text-box: trim-both cap text;
+			text-box: trim-both cap alphabetic;
 		}
 	}
 


johannesmutter:
As a consequence of this change, some gaps and margins may need to be adjusted:

left: after
right: before

The trimming removes the space above letters for more precise control over padding, gap, margin ....

Screenshot 2026-04-29 001203 CleanShot@2x

michael:
Generally I like the behavior of text-box: trim, but I'm not sure if users would expect that to be the default. So from a library point of view I guess it's better to not be opinionated, and let set their text trimming in app-land?

What do you think?

_____

PS: Had `text-box: trim` on for headings in editable website until recently but turned it off, because of that misaligned caret for empty placeholders.

PS2: In this PR the text cursor is still off in placeholders (tested in Chrome 147.0.7727.116)

image

johannesmutter:
I think the expectation from developers is: good spacing out of the box. Not: `text-box` should stay set to its initial value `none`. 

That's why everyone is using Normalize.css or any ui library changes box-sizing to border-box, instead of content-box. Out of historical reasons and because of backwards compatibility some values in css have "outdated defaults". 

I would strongly argue that the same is true for text-box. Even engineers at Google argue that it should be the default, but they just can't make it the default in Chrome, because it will break the web. But for a library that is still in Alpha and that plans to be around for the couple years, I think it's important to be opinionated in regards to getting the defaults right.

Also I'd argue that it's the job of a library to make the DX great.
Having good defaults that I can build on a good design system takes the burden off my shoulders of reinventing _normalise.css for text editors_ everytime I start a new project with svedit

johannesmutter:
> PS: Had `text-box: trim` on for headings in editable website until recently but turned it off, because of that misaligned caret for empty placeholders.
> 
> PS2: In this PR the text cursor is still off in placeholders (tested in Chrome 147.0.7727.116)

Fixed now

johannesmutter:
The design is just more honest with text-box-trim: When I set a padding to 23px, then I expect 23px, not 30px

| text-box: none | text-box: trim cap alphabetic |
|---|---|
| Screenshot 2026-04-30 001214 CleanShot@2x | Screenshot 2026-04-30 001215 CleanShot@2x |

michael:
The question is, is `text-box-trim` the new de-facto standard already, or is it slightly too early? E.g. there still seems to be glitches with it, like the cursor that is off with placeholders (is this a considered a bug I should report at the W3C group?)

Anyways, if we can rule-out surprises/confusion for developers, I'm happy to make text-box-trim the default for Svedit.

My concerns:
- what if i want some other text-box-trim behavior? how can i overwrite? (e.g. in this case i would battle with the !important rules, at least for the placeholders)
- developers assume they just get a browser-native rendering for their AnnotatedText, unless they specify something
- might be more predictable to document a best-practice for the text styles, rather then setting a non-native default, which raises the question, "how should i best unset the default, so i get the native behavior?"

johannesmutter:
I think there are actually two bugs. 
- `trim-both cap alphabetic` doesn't work on the placeholder pseudo element
- caret is not vertically aligned when set on an empty contenteditable element. This is even more noticeable with text-box trim, but exists independent of this new feature. We had in system606, too see `Text.svelte`. The caret seems to use the top edge for placement computation, not the center point between top and bottom, therefore it shifts downwards. Issue is known for 11 years now: https://stackoverflow.com/questions/32905957/caret-position-when-centering-with-flexbox-in-contenteditable/


https://github.com/user-attachments/assets/dbf85eb7-941d-4a49-8060-3012f0fd1fc6

johannesmutter:
> The question is, is text-box-trim the new de-facto standard already

In practice, yes. Any modern UI library that takes spacing seriously has or will adopted it as a default. Adam described it as a recommended baseline at last year’s CSS conference in Amsterdam, and outlined the reasoning clearly in his article: https://developer.chrome.com/blog/css-text-box-trim#feature_and_syntax.

> there still seems to be glitches with it, like the cursor that is off with placeholders

It’s important to separate concerns here: there are no known glitches caused by `text-box-trim` itself. The cursor issue you’re referring to is unrelated to this feature.

> what if i want some other text-box-trim behavior? how can i overwrite? (e.g. in this case i would battle with the !important rules, at least for the placeholders)

You can override it without fighting `!important`. The use of `:where()` intentionally keeps specificity low, so a simple class selector is enough. The only `!important` rule applies to the empty state to stabilize caret placement, it doesn’t affect the rendered placeholder or visual styling. No side effects at all.

> developers assume they just get a browser-native rendering for their AnnotatedText, unless they specify something

Developers generally expect things to work out of the box, without having to patch long-standing browser inconsistencies. Just like we expect `box-sizing: border-box` to be set, it’s reasonable to expect sensible defaults for text layout. If those are missing, it reflects on the library, not the developer.

> might be more predictable to document a best-practice for the text styles, rather then setting a non-native default, which raises the question, "how should i best unset the default, so i get the native behavior?"

Documentation alone doesn’t solve a decades-old problem. Pushing that responsibility onto developers just perpetuates inconsistency. The better question is: why preserve a flawed default at all? Establishing a reliable baseline removes friction and leads to more predictable, consistent UI across the board.

michael:
Completely agree with your points, and I'd love to have text-box-trim on everywhere.

My point is entirely about DX ergonomics and principle of least surprise. Rich text libraries are notorious for setting some styles internally, and often leave the developers debugging for hours until they find out "ahh because [insert rich text library] sets this style rule here. I want Svedit to be as close as possible to to:

```html
<div class="my own classes">My own text</div>
```

so no surprises. Close to full control about rendering without having to sniff into internals.

I'm worried about this scenario:

- We make `text-box: trim-both cap alphabetic;` the default for AnnotatedText (makes sense)
- user wants to change it to say `text-box: trim-both ex alphabetic;` for a specific headline text style
- user sets the style in their own component, which works just fine because of the `:where(...)`
- of course, user doesn't think about placeholders, so the `min-height: 1cap !important;` will cause placeholders to render differently than non-empty text.


That's the kind of surprises I'd like to avoid. You'd not have them if we just relied on the browser defaults (where placeholders work).

Can we somehow get the placeholder pseudo element to inherit the styles and implicit line-height from the main text element, without requiring the user to monkey-patch placeholders as well?

> It’s important to separate concerns here: there are no known glitches caused by text-box-trim itself. The cursor issue you’re referring to is unrelated to this feature.

Yes, it's separate issues, but every new browser feature that works correctly in view-mode but breaks in subtle ways contenteditable=true mode is a problem for us unfortunately. I know there's many weird glitches there, which is exactly why we should clean the mine-field for users by using/promoting only stuff that doesnt cause side effects. 

At the same time I'm really pushing hard in the W3C group to get all the placeholder rendering issues fixed across browsers, and i will push for this one too, to solve the root of the issue.

johannesmutter:
Bug about caret placement is already 12 years old.

https://bugzilla.mozilla.org/show_bug.cgi?id=904846#c46

johannesmutter:
So I have spent now 4hours with claude on trying every possible css attribute combination to fix the caret misalignment in chrome and firefox. display: flex doesn't have the issue. But causes layout shift. Setting the line-height to 1cap also fixes it. But then multiline placeholders have a different appearance as the actual text. Single line placeholders are absolutely fine with this technique.

If we don't want to limit placeholders to single line then the only real way to fix this, besides at the browser level, is to hide the native caret (caret-color: transparent), and to place a css anchor positioned fake blinking caret for when the contenteditable is focused and empty.

**UPDATE:**
We could also accept that on multiline placeholders (anyhow super rare and not recommended) the caret is misaligned and on single-line placeholders we fix it via line-height: 1cap !important;

johannesmutter:
> *I want Svedit to be as close as possible to `<div class="my own classes">My own text</div>`*

Totally agree with that philosophy. But I'd argue `text-box: trim-both cap alphabetic` *is* the closer-to-native behavior in spirit. it's what `<div>My own text</div>` would render as if the CSS spec had been designed in 2026 instead of 1996. The current browser default is the historical accident.

> *user sets the style in their own component, which works just fine because of `:where(...)`. Of course, user doesn't think about placeholders, so the `min-height: 1cap !important` will cause placeholders to render differently than non-empty text.*

I have removed this now. Caret misalignment on empty focused contenteditable is less of an issue than the others in my POV, especially if it's fixable (see above).

> *every new browser feature that works correctly in view-mode but breaks in subtle ways in contenteditable=true mode is a problem*

Agreed in principle, and you're right that we should be careful here. To be concrete though: the caret-in-empty-contenteditable bug isn't caused by `text-box-trim` (the link I posted). `text-box-trim` makes it more *visible* because the box is tighter, but it doesn't introduce it. 

Also we've had this in system606 for 2+ with no reports tied to anything related contenteditable & co. 

So the way I'd frame it: we're not adopting an unstable feature, we're adopting a stable feature that exposes a pre-existing caret bug, which we can now also fix in this PR or another PR. The “mine-field” is smaller after this PR than before it.

> *I'm really pushing hard in the W3C group to get all the placeholder rendering issues fixed*

👍

michael:
Multi-line placeholders are important for me. And I want placeholders to flow like it would flow if it was a real body of text. So I think keeping the caret native is important.

Moving forward I think we could do the change and make `text-box: trim-both cap alphabetic;` the default.

In addition apply the hack in app-land (or svedit-demo-land), to fix the placeholders that are off.

e.g. somewhere in app.css we could add that:

```css
.text.editable.empty) {
  line-height: 1cap !important;
  min-height: 1cap !important;
}
```

just so our demo feels better, and we can do the same in our apps. Meanwhile we push for the issues to be fixed by browsers.

For users of Svedit this means, by default they get `text-box: trim-both cap alphabetic;` but no hacks to fix the browser glitches. They are free to apply the same hack as we do, but then they're aware of it, and own that hack, instead of the library doing magic things.

johannesmutter:
> Multi-line placeholders are important for me. And I want placeholders to flow like it would flow if it was a real body of text. So I think keeping the caret native is important.

Where do you see a relation between placeholders and caret rendering?
Placeholders could be multiline and flow as if the were real text.
And at the same time the caret could be virtual: to also fix the misplacement issues in firefox.

michael:
If you position the caret manually, you have to consider all kinds of things like if the text is left-aligned/centered/right-aligned, potentially margins/paddings etc. We don't know what CSS the user is applying to the text and how the text is rendered, so we don't know where the caret should go, unless we start owning the layouting.

I really think we should be able to rely on native behavior here (unless we take over all rendering one day with pretext et al; not on the table atm) and keep pushing the browser vendors to fix the placeholder rendering.

If we can't find a reliable hack, we might also just accept that FF is not rendering correctly.

johannesmutter:
> … you have to consider all kinds of things like if the text is left-aligned/centered/right-aligned, potentially margins/paddings etc

Not really. You add an inline `<span>` at the beginning of the properties inner content that inherits its height from the font-size multiplied by the caret scale factor, make it 1px wide and reuse the existing node-gap caret blinking animation.  Then you hide the native caret. That's it. It will always render at the start of the placeholder.

I'm not arguing to replace the native caret fully. I'm arguing to fix an issue that browser vendors didn't care to fix for 12+ years, with a couple lines of css and one line of html. The caret for the placeholder needs no logic, it just sits at the beginning of the placeholder.

I might also miss something.

johannesmutter:
not even a span is necessary. 9 lines of css: problem solved!

https://github.com/user-attachments/assets/c729078c-404a-4fc7-bdf1-37fd495166ba

The css (needs to be optimised to precisely match real caret):

```css
[placeholder].editable.empty.focused:before {
    content: "";
    width: 0px;
    height: 1em;
    background: blue;
    display: inline-block;
    box-shadow: 0 0 0 0.7px blue;
    animation: var(--node-caret-animation, svelte-s92v36-node-caret-blink var(--node-caret-blink-duration, 1.1s) ease-in-out infinite);
    transform: translateY(4px);
}
```

michael:
I'd expect the caret to be centered here:

image

But happy to experiment with it. Maybe in a separate PR?

michael:
One could also decide to render the caret at the beginning even when centered, but downside: As you start typing, your caret jumps quite far. Like you type one character at the top left, and your new character appears somewhere else.

johannesmutter:
> I'd expect the caret to be centered here:
> 
> image
> But happy to experiment with it. Maybe in a separate PR?

hmmm I think this looks broken when the caret is rendered on top and in the middle of a character.
Also it's unexpected: users tend to click towards the beginning or end of a text (not middle), even for inputs with placeholders. If the caret is then rendered in the middle you also have that jump from where you click to now the center point.

It feels especially strange when you go from a previous node via arrow keys to the next node, and then for center aligned text only the caret is not at the start of the text. Inconsistent I'd say

And yet another case is pasting text: also here it would be a suprise if the caret was in the middle

michael:
Me:

> One could also decide to render the caret at the beginning even when centered, but downside: As you start typing, your caret jumps quite far. Like you type one character at the top left, and your new character appears somewhere else.

GPT 5.5

image

Opus 4.7

image

johannesmutter:
are we now outsourcing our thinking ?

Screenshot 2026-05-01 001231 Claude@2x

michael:
Lol, no. Shared those outputs, as they seem to confirm that "caret should be where the first character will land" is what's the most popular interpretation out there.

But the better argument to "this seems to be the default currently" is looking how native text input and text area behave with caret placement.

Here's an isolated example, showing text area, input and contenteditable.

[caret-in-centered-text.html](https://github.com/user-attachments/files/27272872/caret-in-centered-text.html)

image

For textarea/input they caret renders centered in all browsers. For the contenteditable version, Safari/Chrome do too. Just Firefox messes things up here, and renders to the very left. Bug is known (see https://bugzilla.mozilla.org/show_bug.cgi?id=2005674, https://github.com/w3c/editing/issues/501)

___

That said, it's fine to question and challenge this behavior, so good to have #258 on the table. Will do more testing, maybe this could be a temporary (until FF fixes the bug) or permanent solution (if we decide that cursor placement at start of placeholder feels better)


___

As for this PR, I'd suggest we switch to `text-box: trim-both cap alphabetic;` as the default, but document this choice in the README and also add a section "Recommended workarounds" to the README, to give people strategies to "fix broken browser things at their own risk".

Something like this:

```css
/*
  Fix vertical alignment of caret due to a decade-old 
  https://stackoverflow.com/questions/32905957/caret-position-when-centering-with-flexbox-in-contenteditable/

  This is safe to use when your app/website only uses `text-box: trim-both cap
  alphabetic;` (Svedit default) for all `<AnnotatedTextProperty>` istances. If you want
  to use other text-box trim strategies, you have to make specific overrides to
  .text.editable.empty. for each case.
*/
.svedit-canvas .text.editable.empty) {
  line-height: 1cap !important;
  min-height: 1cap !important;
}
```

Does that make sense?

johannesmutter:
I want to gently push back on the form of reasoning here. "What browser vendors do," "what another text editing library does", "what AI says without prompting for critical reflection" … these are useful signals about **what's common**, but I don't think they're a strong basis for deciding **what's right**. Popular and correct aren't the same thing, and design decisions in particular deserve to be re-examined on their own terms.

What I find more reliable is to look at a problem with fresh eyes (& thinking from first principles): setting aside how it's usually done, setting aside one's own preconceptions shaped by years of using existing tools, and walking through all the scenarios a user actually runs into. Not just one canonical action like clicking, but arrow-key navigation, paste, the different alignments, RTL, vertical writing, and aesthetics. That's where trade-offs surface that the popular default tends to hide.

I'd love it if we could approach this and other such question that way too. Even when the precedent comes from a browser vendor or someone with real authority in the space. Not to be contrarian, but because looking at it fresh is the only way to find out whether the popular answer also happens to be the right one.

And if the answer for yourself then still is: "this doesn't feel right, here's why ...", and mine is the opposite, then it's maybe a signal to do a carefully prepared lab experiment and test with other humans.

---

> As for this PR, I'd suggest we switch to text-box: trim-both cap alphabetic; as the default, but document this choice in the README and also add a section "Recommended workarounds" to the README, to give people strategies to "fix broken browser things at their own risk".

Sounds good. I'll add the documentation now

johannesmutter:
When looking at what others do, I think it's important to filter out what feels right and what doesn't feel right to **your specific context**:

https://github.com/user-attachments/assets/595eb7b2-5d69-4c2c-bf7d-c66a6e1f19d7

In Pages the placeholder is actual content, therefore the caret is placed at the end when you click at the end: the intention is correctly interpreted: I want to place the caret at the end of this existing text.
When you click somewhere before the end: the entire text is selected so you can easily override with just one keystroke.

https://github.com/user-attachments/assets/a981336e-c402-4a92-a384-280e3f3be7ca

In keynote the placeholder disappears immediately on focus. That feels even better to me, as a placeholder is a distraction to my thinking when I'm about to write something. 

In browser form inputs the placeholder stays because form inputs are different: you need to keep the instructions visible during the thinking process: "add your telephone number here". This is a very different type of placeholder as in our text editor: "Write a paragraph", "This is a heading". These placeholders are not about "required" and "semantically shaped" form input, but just serve as a visual cue: this is what a paragraph looks like. When you just look at how browser form inputs handle this, then you fall into that trap of missing what's actually relevant here.

johannesmutter:
> document this choice in the README and also add a section "Recommended workarounds" to the README, to give people strategies to "fix broken browser things at their own risk".

This is now documented. See this PR https://github.com/michael/svedit/pull/261

PR #254, opened 2026-05-01
Also create a node payload for single paragraphs at a node gap

johannesmutter:


michael:
Can you add exact steps to reproduce the problem this PR fixes?

johannesmutter:
1. Copy this text:
```Hello World```
2. Paste here: https://svedit.dev/ in a NodeGap
3. Nothing happens

---

1. Copy this text:
```Hello World```
2. Paste here: https://svedit-opvj17io2-michael-aufreiters-projects.vercel.app/ in a NodeGap
3. Text is inserted

**Comment by johannesmutter** [comment]
The bug: Pasting a single paragraph of plain text at a node gap does nothing — the paste is silently swallowed.

Why it happens: The paste handler splits plain text by \n\n. With a single paragraph, you get 1 fragment. The old condition plain_text_fragments.length > 1 is false, so no pasted_json is created. The dispatch then falls through to the last resort:

```javascript
session.apply(session.tr.insert_text(plain_text.trim()));
```

But insert_text bails immediately when selection.type !== 'text' (line 661 of Transaction.svelte.js). At a node gap, the selection type is 'node'. So the paste is dropped.

Multi-paragraph text worked fine because length > 1 created a pasted_json, which matched the pasted_json?.main_nodes && session.selection?.type === 'node' condition and routed through try_node_paste.

PR #253, opened 2026-04-30
Serialize/deserialize document paths reliably

michael:
Switch path handling to serialized "__"-separated strings for data-path, anchor names, and position anchors. Add serialize_path/deserialize_path and path-segment validation, use paths_equal where appropriate, migrate mounted_paths to a reactive record, update node-gap parsing, and add tests for path serialization and node id validation.

PR #255, opened 2026-05-01
Toggle IO visibility classes on nodes based on existing intersection observer infrastructure

johannesmutter:
### What's in this PR

**1. New feature: viewport visibility classes** (opt-out via `session.config.view_classes`)

Each `[data-type="node"]` element gets `.in-view` / `.seen` / `.fully-in-view` / `.visible-top` / `.visible-bottom` toggled as it scrolls through the viewport. Useful for scrollytelling, animations, performance optimisations e.g. hiding expensive components when out of view, stopping playback for videos that are out of view, etc.

On by default — costs ~5 classList ops per IO entry.

Class semantics:
- `.in-view` — any part of the node intersects the viewport
- `.seen` — set once, never removed (appeared at least once)
- `.fully-in-view` — neither top nor bottom edge is clipped
- `.visible-top` — top part showing, bottom clipped below viewport
- `.visible-bottom` — bottom part showing, top clipped above viewport
- The three positional classes are mutually exclusive (a node taller than the viewport gets none of them — just `.in-view`)

**2. Rewrite: `node_gap_computation.svelte.js` (850 lines) → `node_visibility.svelte.js` (~480 lines)**

Two `IntersectionObserver`s + one `MutationObserver`. Replaces the previous dual-observer + hand-rolled gap cache + per-array signal pipeline with native `SvelteMap`/`SvelteSet` reactivity for marker rendering and an imperative `classList.toggle` path for `.positioned` activation.

- **Overscan IO** (`rootMargin: 500px`, threshold `[0]`) — drives `near_map`, `clip_map`, `array_indices`, and imperative `.positioned` toggling on adjacent gaps
- **Viewport IO** (`rootMargin: 0`, threshold `[0, 1]`) — drives view classes only. A separate observer is required because the overscan IO fires its thresholds while nodes are still 500px outside the real viewport — by the time a node scrolls into view, both thresholds have already been crossed and no further callback fires
- `near_map` (SvelteMap) — per-node "in overscan zone" flag
- `clip_map` (SvelteMap) — per-node edge-clip state for nested scrollers
- `array_indices` (Map\<path, SvelteSet\>) — per-array set of near child indices, used by `NodeGapMarkers` to iterate only visible offsets (O(K) instead of O(N))

**3. Race-condition fix for newly inserted nodes**

When a node is inserted into the DOM, the MO callback unobserves it from the viewport IO, queues a `requestAnimationFrame`, then in the RAF applies view classes via `getBoundingClientRect` and re-observes. This prevents the viewport IO's initial callback (which fires in a subsequent frame with potentially stale intersection data) from undoing the correct classes.

**4. Perftest route** (`/perftest`) — mount, scroll, resize, and mutation FPS benchmarks at 50–2000 nodes with mixed node types.

### Performance

Mount time at 2000 nodes: **3.7s → 2.3s** (-38%).

The big win came from a DevTools Performance trace that attributed 1.5s of forced reflow to `getBoundingClientRect` calls in the bootstrap. Those calls were interleaved with Svelte's mount work, so each gBCR re-triggered layout for the growing DOM. Removed them entirely — IO populates the maps on its first callback, ~1 frame after `observe()`. Un-positioned NodeGaps have zero layout presence so the lag is invisible.

| nodes | mount (ms) | scroll avg (fps) | resize avg (fps) | mutation avg (fps) | memory (MB) |
|------:|-----------:|------------------:|------------------:|-------------------:|------------:|
|    50 |         46 |                60 |                60 |                 34 |          31 |
|   100 |         74 |                58 |                60 |                 34 |          34 |
|   200 |        102 |                51 |                57 |                 21 |          49 |
|   500 |        297 |                25 |                18 |                 10 |          53 |
|  1000 |        755 |                 7 |                 4 |                  3 |          90 |
|  2000 |       2344 |                 2 |                 1 |                  1 |         204 |

The 50–500 range covers the realistic editing scale and is smooth (≥25 fps scroll, sub-second mount). 1000+ nodes degrade — primarily browser-level layout cost over ~13K DOM elements that we can't optimize from JS.

### Tradeoffs and things rejected

- **Single IntersectionObserver for both overscan and viewport** — a single observer with `rootMargin: 500px` can't reliably track actual viewport entry. For nodes shorter than 500px, both thresholds are crossed while the node is still in the overscan buffer, so no callback fires when it enters the real viewport. Benchmarks confirmed two observers have no measurable FPS impact.
- **`content-visibility: auto` on `.node`** — broke multi-node selection across viewports; selection overlays disappeared when scrolling back over previously selected (now culled) ranges. NodeGap rendering also became inconsistent. Documented as not viable for an editor.
- **Lazy NodeGap rendering** — would break the "always present in DOM" rule that selection anchors and scrollTo depend on.

### Architecture notes

- `.positioned` activation on NodeGap is **imperative** (classList.toggle from the IO callback), not a reactive `$derived`. At 1000+ NodeGaps the per-instance signal allocation was measurable; the imperative path uses sibling traversal on the IO entry's target element to find adjacent gaps. Trade-off: a small DOM coupling, but it eliminates ~3000 reactive signals at scale.
- **`NodeGapMarkers` iterates `visible` indices**, not `0..count`. At 1000 nodes with ~30 visible, that's 30 iterations per re-derive instead of 1001.
- **Bootstrap does NOT call `getBoundingClientRect`** — IO populates state on its first callback. The MO callback does use BCR for newly inserted nodes, but batched into a single RAF to avoid layout thrashing.
- **MO callback is a single pass** over mutations, handling both node observation and gap `.positioned` sync in one iteration.

---

Summary written by Claude


johannesmutter:
Full rewrite of node visibility tool complete.
Improved performance with 500 nodes now running smooth 30fps, using more realistic tests (e.g. realistic scrolling speed). Also includes performance test for mutations. 


## Claude highlighted these issues @michael :

**1. MO amplification from `(index)` keying**

`NodeArrayProperty.svelte` uses `{#each nodes as node, index (index)}` — keyed by *position*, not by node identity. When a button is inserted at position 0, every subsequent node shifts index. Svelte sees key 0 change from node A to node B, key 1 change from B to C, etc. — it destroys and recreates *every shifted component*. The MO fires for all those removals + additions.

This means inserting 1 node into an array of 20 triggers ~20 component destroy/create cycles, ~20 MO removal events, ~20 MO addition events, plus all the `observe`/`unobserve`/`schedule_view_sync` calls for each.

Fixing this: use the **node ID** as the each-block key (`{#each nodes as node, index (node.id)}`). Svelte would then recognize that existing nodes just moved, reuse their DOM elements, and only create the one new component. The MO would fire for 1 addition instead of 20.

**2. Mutation FPS is dominated by Svelte transaction cost**

The benchmark showed ~5-6 FPS for single-node-per-frame mutations even at 50 nodes. The visibility system (IO, MO, BCR) adds negligible overhead — it's the `session.tr` → `structuredClone` → reactive state update → Svelte reconciliation chain that costs ~170ms per frame. That's a Session/Transaction layer concern, not a visibility one.

**Proposed things to fix :**

- **`NodeArrayProperty.svelte` line 69**: change `(index)` to `(node.id)` to eliminate the MO amplification. This is the single biggest win for mutation performance. It would also reduce unnecessary `schedule_view_sync` calls since existing nodes wouldn't be re-added to the DOM.
- **Transaction/Session layer**: the `structuredClone` on every `session.apply()` and the reactive state propagation are the real bottleneck for mutation FPS at scale.

---

I remember you said using the node.id is not feasible? But can't recall where.

johannesmutter:
**Heads-up for whoever merges this after #253 (path-serialization) lands**
#253 changes the canonical path-string separator from `.` to `__` (via `serialize_path()` in `src/lib/utils.js`, where `PATH_SEPARATOR = '__'`). This PR's `node_visibility.svelte.js` is written against the old `.` convention — once both are on main, no `.positioned` class ever gets toggled and `array_indices` is never populated, so **all NodeGaps stay invisible and no NodeGapMarkers render at all**.
Two specific things break:
1. `should_position_gap_imperative` builds lookup keys as `` `${array_path_str}.${i}` ``, but `near_map` / `clip_map` are keyed off `el.dataset.path`, which #253 emits via `serialize_path()` (e.g. `body__0`). Lookups never hit → no `.positioned` toggle → zero-size invisible gaps.
2. `#split_path` does `path.lastIndexOf('.')` and returns `null` for `__`-separated paths, so `array_indices` (which `NodeGapMarkers` iterates) stays empty.
### Required changes when merging
First, export `PATH_SEPARATOR` from `src/lib/utils.js` (one-liner: add `export` to the existing `const PATH_SEPARATOR = '__'`).
Then in `src/lib/node_visibility.svelte.js`:
```js
import { PATH_SEPARATOR } from './utils.js';
// #split_path
const sep = path.lastIndexOf(PATH_SEPARATOR);
if (sep < 0) return null;
const index = parseInt(path.slice(sep + PATH_SEPARATOR.length), 10);
if (Number.isNaN(index)) return null;
return { array_path: path.slice(0, sep), index };
// should_position_gap_imperative — replace every `${array_path_str}.${i}`
// with `${array_path_str}${PATH_SEPARATOR}${i}` (5 occurrences).
```
### Related fix outside this PR
`src/lib/NodeGapMarkers.svelte` (already on main) has the same problem — it builds anchor names with `path.join('-')` / `${anchor_prefix}-${i}` and map keys with `${path_str}.${i}`. After #253 lands, those need to switch to `serialize_path(path)` / `${PATH_SEPARATOR}${i}` to match the anchor names emitted by `Node.svelte` and `NodeGap.svelte`. Worth landing as a follow-up commit on main alongside #253, otherwise even after fixing this PR the markers stay broken.

johannesmutter:
## Benchmark Results (500 text nodes, edit mode)

### Microbenchmarks: raw operation cost

| Batch size | classList (µs) | SvelteSet (µs) | SvelteSet steady-state (µs) | Ratio |
|---|---|---|---|---|
| 5 | 1.8 | 0.8 | 0.4 | 0.4x |
| 10 | 2.5 | 0.3 | 0.2 | 0.1x |
| 15 | 9.6 | 2.8 | 2.0 | 0.3x |
| 25 | 6.1 | 0.5 | 0.4 | 0.1x |

**SvelteSet writes are 3–14× faster** than classList toggles in isolation.

### Scroll FPS: system-level impact

Without reactive subscriptions (previous test, standard Text component):

| Mode | Avg FPS | Min FPS |
|---|---|---|
| baseline (no subs) | 40 | 20–29 |
| + classList extra | 39–41 | 20–29 |
| + SvelteSet (no subs) | 40–43 | 20–30 |

With 500 `$derived(set.has(path))` subscriptions (TextWithVisibility):

| Mode | Avg FPS | Min FPS |
|---|---|---|
| baseline (subs idle) | 32 | 20 |
| + reactive subs active | 29–31 | 12–20 |
| + classList extra | 34 | 20–29 |
| + SvelteSet (no subs) | 34–37 | 20–30 |

### Key finding

**500 idle `$derived` subscriptions to a SvelteSet cost ~20% FPS** (40 → 32).
Actually triggering them adds only a small further cost (32 → 29–31).
The overhead is in the reactivity graph structure, not the per-mutation work.

Additionally: initial load shows visible layout jumping (3 reflows before settling)
when 500 components each hold reactive subscriptions that interact with the
gap-positioning IO cycle.

michael:
#253 is now merged. Green light for getting this one ready.

michael:
Synced with main now, according to instructions. No issues from my testing, though I'm not sure I'm looking at all the right things.

@johannesmutter can you do a final check if everything is alright and this is ready for merge?

PS: let me push the merge button pls, want to do a bit more testing with the editable website first, if some issues come up.

johannesmutter:
tested node selections by keyboard and mouse. tested with 2000 nodes: copy, cut, paste, move, undo & redo, scroll out of view, scroll back to view. ran performance tests.

Results:

| 50    | 68        | 673    | 21   | 31     | 60         | 56         | 60         | 57         | 32           | 20           |
| 100   | 74        | 1353   | 19   | 30     | 58         | 29         | 60         | 56         | 29           | 20           |
| 200   | 134       | 2715   | 18   | 32     | 41         | 20         | 53         | 29         | 22           | 15           |
| 500   | 305       | 6825   | 18   | 52     | 19         | 12         | 17         | 9          | 8            | 7            |
| 1000  | 719       | 13681  | 17   | 113    | 7          | 3          | 4          | 2          | 3            | 3            |
| 2000  | 2404      | 27374  | 17   | 171    | 1          | 1          | 1          | 1          | 0            | 0            |


No issues noticed from my side.

michael:
Looks also good inside editable-website integration, and on mobile.

There's one issue (also present in main, but not in v.0.9.0) where anchor positioning at the overlay layer doesn't work as expected for scrollable containers.

It seems the popover's target is now the whole container, not the focused nav item node anymore. Scrolling horizontally doesn't move the popover.

https://github.com/user-attachments/assets/ae23b4b8-1d5b-4fce-b850-278b62344f6e

In 0.9.0 this part works (but there node gaps do not reposition as you scroll - only after scroll stopped):

https://github.com/user-attachments/assets/1508b957-7b0d-49d7-a191-26ce7ece99d4

michael:
Merging already, created a new issue #268 for the anchor positioning problem. @johannesmutter maybe you can take a look there. That's from my side probably the only blocker remaining for a 0.10 release.

johannesmutter:
> It seems the popover's target is now the whole container, not the focused nav item node anymore. Scrolling horizontally doesn't move the popover.

Not sure need to look into this again, but I think this is intentional, so that the popover doesn't get cropped, but stays in the visible viewport

michael:
I guess they should stay in viewport of the main window (not the scrollable container)? And that may be a concern of each popup indvidually? Like how it finds it's optimal position, not covering the anchor etc.


CreateLink.svelte in editable website currently has this markup, and it's clamped to the viewport generally.

```svelte
<dialog
	bind:this={dialog_ref}
	class="create-link-dialog absolute z-40 mt-1 m-0 bg-(--background) text-(--foreground) p-0 shadow-xl overflow-visible max-h-90 border border-[color-mix(in_oklch,var(--foreground)_18%,transparent)]"
	style="position-anchor: --selection-highlight; position-area: block-end span-all; justify-self: anchor-center;"
	onclick={handle_backdrop_click}
>.../dialog>
```

Centered to the anchor when possible, but clamped when not:

image

Easiest to test I guess is with editable-website#main against svedit#main with a sveltekit alias, but let me know if we should work on a setup/example that lets us reproduce/test this in svedit itself?

johannesmutter:
I think we should have a minimal popover toolbar e.g. url editor (instead of the current native window modal) so we can test in svedit itself, also with a test where we measure the location within a vertical and horizontal scrollable container against expected placement.

michael:
Would be good to setup something that can become a permanent part of the demo, so we can always verify if the popover positioning works in combination with overflowing containers.

Here's a starting point: https://github.com/michael/svedit/pull/270

PR #256, opened 2026-05-01
Remove arbitrary parent traversal limit

johannesmutter:
## Problem

When pressing ESC repeatedly to climb the document hierarchy, the selection would abruptly disappear at certain nesting levels instead of continuing to climb.

The root cause: `select_parent` navigated upward by slicing 2 segments off the selection path on each call (`['page_1', 'body', 3, 'buttons', 1, 'label']` → `['page_1', 'body', 3, 'buttons']` → `['page_1', 'body']`). But it only did this when `path.length > 3`. Once the path reached length 2 — e.g. `['page_1', 'body']`, meaning "a node selection on page_1's body array" — it hit the `else` branch and cleared the selection entirely.

The issue is that at length 2, the path no longer encodes who the *parent* of `page_1` is. The old code had no way to answer "which array contains `page_1`?" so it gave up. This meant the last possible climb (from a top-level array up to the node that owns it within *its* parent array) was never reachable.

## Fix

When the path is too short to encode the parent (length < 4), walk the document graph (`_find_node_parent`) to find which node holds the subject in one of its `node_array` properties. This resolves the final hop. If no parent exists (i.e. we've reached the document root), selection is cleared as before.

## How to test

1. Open a document with deeply nested node arrays
2. Click into text inside the nested content
3. Press ESC repeatedly — selection should climb through each level of the hierarchy before finally clearing at the root

### DIFF
--- a/src/lib/Command.svelte.js
+++ b/src/lib/Command.svelte.js
@@ -90,14 +90,16 @@ export class RedoCommand extends Command {
 /**
  * Command that selects the parent of the current selection.
  * Useful for navigating up the document hierarchy.
+ *
+ * Enabled whenever there is any selection — `select_parent()` walks the
+ * document graph to find the containing array, so paths that don't encode
+ * the parent (every node selection, top-level text selections) still
+ * climb correctly. The previous `path.length > 3` gate was a string-only
+ * heuristic that silently disabled climbing for nested arrays.
  */
 export class SelectParentCommand extends Command {
 	is_enabled() {
-		return (
-			this.context.editable &&
-			this.context.session.selection &&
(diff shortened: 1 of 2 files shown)


michael:
It's a choice to be questioned, but I disabled selecting parents because I thought you'd never want to lose the selection completely if there's no selectable parent.

This commits changes the select parent behavior to stop at the last selectable parent: https://github.com/michael/svedit/commit/213126a5660476c22109ae91fe6a392ee87b8970

johannesmutter:
I just compared the same deeply nested components on this branch vs. current state on main. 
I'm no longer experiencing this issue. 
Closing now.

PR #257, opened 2026-05-01
Fix caret jumping to offset 0 when clicking into text

johannesmutter:
## Summary
- Removes `removeAllRanges()` from `render_selection()` when `session.selection` is null
- Fixes caret being placed in front of the first letter instead of at the click position when clicking a text property from outside the canvas

## Root cause
`handle_canvas_focus` sets `session.selection = null` and flushes synchronously. The `$effect` for `render_selection` fires during `flushSync`, sees no selection, and calls `removeAllRanges()` — wiping the browser's pending click selection. The browser falls back to offset 0.

## Test plan
- [ ] Click into a text property from outside the canvas, verify caret lands at click position
- [ ] Click into a text property that has a visible (blurred) selection highlight, verify caret lands correctly
- [ ] Verify existing selection behavior (arrow keys, shift-select, etc.) is unaffected

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -703,9 +703,6 @@ ${fallback_html}`;
 		const selection = /** @type {Selection} */ (session.selection);
 
 		if (!selection) {
-			// No model selection -> just leave things as they are
-			let dom_selection = window.getSelection();
-			dom_selection.removeAllRanges();
 			return;
 		}
 


michael:
Regarding this one: Yes this is some regression introduced with https://github.com/michael/svedit/commit/c5663fccf3f78387b2094e5b48d9f83c7431d6ba.

Need to double-check that your fix doesn't regress somewhere else, but it's looking good from a first look.

michael:
Merging this now at the risk of reverting the change later, if it introduces new problems (can't see any now)

PR #258, opened 2026-05-01
Fix caret vertical alignment in empty text properties with placeholders

johannesmutter:
## Problem
When focusing an empty text property that has a placeholder, the native caret in Chrome and Firefox is vertically misaligned — it doesn't align with the placeholder text rendered via `::after`.

## Solution
- Hide the native caret on empty focused editable text properties with placeholders
- Render a virtual caret via a `::before` pseudo-element using box-shadow, sized to `1cap` with vertical extensions to match the text line
- Reuses the node caret's blink animation and respects `--svedit-caret-color`

## Test plan
- [x] Focus an empty text property with a placeholder — caret should be vertically aligned with the placeholder text
- [x] Verify on Chrome, Firefox, and Safari

### DIFF
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -167,6 +167,27 @@
 		color: color-mix(in oklch, currentcolor 50%, transparent);
 	}
 
+	/* A virtual caret: to fix the caret vertical alignment issue in Chrome and Firefox for empty focused contenteditable with placeholders */
+	[placeholder].editable.empty.focused  {
+		caret-color: transparent;
+	}
+	[placeholder].editable.empty.focused::before {
+    content: "";
+		/* we limit width & height to avoid layout shifts in case the text has a lower natural height */
+    width: 0px;
+    height: 1cap;
+    display: inline-block;
+		/* we use box-shadow to draw the caret shape, matching the native caret */
+    box-shadow: 
+			0 -0.4cap 0 0.65px var(--svedit-caret-color, AccentColor),
+			0 0 0 0.65px var(--svedit-caret-color, AccentColor),
+			0 0.4cap 0 0.65px var(--svedit-caret-color, AccentColor);
+		animation: var(
+			--node-caret-animation,
+			node-caret-blink var(--node-caret-blink-duration, 1.1s) ease-in-out infinite
+		);
+	}
+
 	/* Hide flickering: in Chrome, the caret jumps from end of placeholder string to start of text property when we focus */
 	.text:not(.focused) {
 		caret-color: transparent;
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -168,7 +168,7 @@
 	}
 
 	/* A virtual caret: to fix the caret vertical alignment issue in Chrome and Firefox for empty focused contenteditable with placeholders */
-	[placeholder].editable.empty.focused  {
+	.empty  {
 		caret-color: transparent;
 	}
 	[placeholder].editable.empty.focused::before {
--- a/src/lib/AnnotatedTextProperty.svelte
+++ b/src/lib/AnnotatedTextProperty.svelte
@@ -168,8 +168,10 @@
 	}
 
 	/* A virtual caret: to fix the caret vertical alignment issue in Chrome and Firefox for empty focused contenteditable with placeholders */
-	.empty  {
-		caret-color: red !important;
+    /* Browser BUG: iOS Safari only considers the caret color set on the top contenteditable element, not on nested elements (e.g. the second selector doesn't work in iOS Safari) */
+	:global(.svedit-canvas:has([placeholder].editable.empty.focused)), 
+    [placeholder].editable.empty.focused {
+		caret-color: transparent !important;
 	}
 	[placeholder].editable.empty.focused::before {
     content: "";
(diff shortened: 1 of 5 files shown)


michael:
For empty text properties we'd need to hide the native cursor, otherwise we get 2.

image

johannesmutter:
> For empty text properties we'd need to hide the native cursor, otherwise we get 2.

This looks like an issue in iOS Safari only. maybe they don't support caret-color?

johannesmutter:
Fixed: iOS Safari has a browser bug. caret-color is only considered from either only the upmost contenteditable element or doesn't support nested overrides of caret-color.

michael:
Preferring to get this merged as it:

- addresses all the current browser issues, including Firefox caret rendering
- might allow us to remove the conditional `<br>` removal on empty fields (less workarounds needed, will test this)

Left to check/verify:

- [x] Will this also work when no placeholder is used/rendered?
- [ ] If no, should we render the native caret in that situation, or work towards making the custom caret work without placeholders too?

I added a native-first (when possible) principle to the Svedit README (see [here](https://github.com/michael/svedit#principles)). The reason is that I'd like to see most of Svedit disappear eventually, and be replaced native behavior that 'just works'. 

Yes there are currently a lot of deficiencies and decade-old bugs, but that's because caret movement has never been standardized. But this changes now with the new [Caret Movement Spec](https://github.com/w3c/editing/issues/529). I'm part of drafting the initial spec, and would like to use that opportunity to fix and improve things at the heart.

Regarding the question whether by default the caret should be placed at beginning of a placeholder instead of at the insersition pos of the next character, I'm undecided personally, but would be happy to bring this to discussion at the Editing Working Group:

Can you write up a short paragraph outlining why this is better, and why the drawbacks (cursor might land far away after the typing the first character - in particular for right aligned text, where the cursor might jump all the way from the left edge to the right) are justified?


PS: We could also open discussions around cursor affinity (maybe we can get a native API for that) among other things.

johannesmutter:
> * might allow us to remove the conditional `<br>` removal on empty fields (less workarounds needed, will test this)

The conditional `<br>` removal is no longer happening. It's not added in the first place in NodeArray. Only consistently present in NodeGap as it should be.

johannesmutter:
### Why the caret should always render at the start of the text

The caret in an empty editable element, with or without placeholder, should anchor to the start of the line, not to a position derived from the text frame and its alignment (e.g. center for `text-align: center`). The frame-based approach optimizes for a single moment, the instant before the first keystroke, while the start-based approach holds up across every interaction a user actually performs.

**Concretely:**

- **Click.** Clicking anywhere in an empty centered field places the caret on top of the placeholder text, which reads as broken rather than as a helpful preview. A caret should never render on top of a letter, but only between letters. Readability and aesthetics suffer if a caret cuts a letter visually in half, or lands e.g. exactly on a vertical stroke of a glyph making it invisible even though it's blinking (assuming the caret has the same color as the text).
- **Arrow-key navigation.** Filled blocks land the caret at the start of the text. If empty centered blocks jump to the middle instead, navigation becomes inconsistent across block types. Center aligned text is not not term but the exception.
- **Paste.** Insertions feel detached from where the user thinks they "are" in the document when the caret sits somewhere they didn't click.
- **Writing direction.** "Start" means left in LTR, right in RTL, and top in vertical scripts. Anchoring to the start respects the script automatically, with no special casing for alignment.
- **Visual integrity.** A caret rendered at the start never collides with the placeholder glyphs, so the empty state always looks clean.

The counter-argument, that the caret should preview where the first character will appear, sounds principled but trades consistency in click, navigation, paste, and writing direction for one frame of visual prediction. That trade is not worth it.

---

The Caret doesn't render readable when centered over text

https://github.com/user-attachments/assets/6453638a-8e68-4f0e-aa3e-7cab9abd9448

Scanning happens on the left edge of text, when the caret stays "start of text aligned", scanning is easier, with less jumps and recallibration needed by the eye.

https://github.com/user-attachments/assets/a66f569d-74e2-42d6-8720-91753c45c2ee

johannesmutter:
>  Will this also work when no placeholder is used/rendered?

It works: [see clicking in any paragraph here](https://svedit-git-fix-empty-text-ca-e333bb-michael-aufreiters-projects.vercel.app/) after removing the contents

johannesmutter:
> in particular for right aligned text, where the cursor might jump all the way from the left edge to the right

The jumping of the cursor from the left to the right is actually preferred. If you click inside the text and the caret would render behind the last "e" of the first line it would feel like the computer misinterpreted your intention. 

What's bothering more here is the layout shift because the placeholder is multiline. I think placeholders should ideally be a single line of text.

https://github.com/user-attachments/assets/24a25bcd-c185-417e-a149-6d11b3f9fe9c

michael:
> It works: [see clicking in any paragraph here](https://svedit-git-fix-empty-text-ca-e333bb-michael-aufreiters-projects.vercel.app/) after removing the contents

what i meant is will the custom cursor also render/position correctly if you set `placeholder=""`. I checked, and yes it does.

> The conditional `<br>` removal is no longer happening. It's not added in the first place in NodeArray. Only consistently present in NodeGap as it should be.

Not talking about the node array here, but the br in AnnotatedText.svelte, that is shown/hidden based on wether the field has focus or not (we did this so the native cursor will not be rendered after the placeholder). Trying to get rid of this two-state thing now, and checking if the custom empty-caret still renders correctly.

michael:
Ok, nope, we can't get rid of hiding the br conditionally (when focused or when empty).

Nevermind, the custom cursor works in all situations I tested, across all browsers.

Issue #251, opened 2026-04-29
Conditional `NodeGap` rendering forces every adjacency rule to be written twice


johannesmutter:
`NodeArrayProperty` only renders `<NodeGap>` divs when `svedit.editable === true`. In read mode the gaps don't mount at all.

That means CSS rules that target block-to-block adjacency have to ship in two forms — one walking past the gap, one not:

```css
/* edit mode (gap div is interleaved) */
.node-array > .node-text + .node-gap + .node-media,
/* read mode (no gap div) */
.node-array > .node-text + .node-media {
   margin-block-start: 1rem
}
```

CSS sibling combinators see hidden elements as siblings, so we can't fix this in CSS alone — `display: none` on the gap wouldn't help. Every pair-table rule literally needs to list both selectors. 

In my codebase the pair table doubled in line count and every contributor has to remember the convention or one mode silently has wrong spacing.

### Proposed fixes (any of these would solve it)

1. **Always render `<NodeGap>`, hide it in read mode.** Make NodeGap render with `display: none` when `!svedit.editable`. The DOM stays consistent across modes; consumer CSS only ever needs the `.node-A + .node-gap + .node-B` form. `NodeGap` in read mode can also be a much simpler dom element. just a div with the class name.
2. **Document the dual-form requirement in the API docs / Node component JSDoc** so people targeting adjacency know up front. (Worse than the above two — keeps the footgun.)
3. Find a solution that doesn't mess with the HTML structure.

michael:
Phew, yeah I'm really undecided about this one. It's surprising that there's node gaps rendered as siblings, changing how you can address childs in a node array. It's even more surprising when the gaps are there in edit mode but missing in edit mode.

BTW, on the bright side, turning the node gaps into siblings of the nodes fixed cursor up/down navigation in Firefox (like this issue here https://github.com/w3c/editing/issues/523 is not affecting us anymore, a least not for the node_array node gaps)

As to solve this, I'd say we go for (1) until we find a better solution (3). That said I think we should keep all the markup the same between edit and view mode, and really just add the `display :none`. This way we are just being honest about the infrastructure bits that Svedit adds, and you can rely on a certain structure. If we change markup between edit/view mode we just add more confusion i think.

Wdyt, does that make sense?

johannesmutter:
[...]

PR #260, opened 2026-05-01
Node gaps always present in DOM (viewing and editable)

johannesmutter:
with documentation and example in List component for new css selectors;

Potentially breaking rename:
"empty-node-array" → "empty-node-placeholder"

Fixes this issue:
https://github.com/michael/svedit/issues/251#issuecomment-4356400668

### DIFF
--- a/README.md
+++ b/README.md
@@ -1087,8 +1087,59 @@ Further things to consider:
 - Never apply `position: relative` to the direct parent of `<AnnotatedTextProperty>`, it will cause a [weird Safari bug](https://bsky.app/profile/michaelaufreiter.com/post/3lxvdqyxc622s) to destroy the DOM.
 - Never apply `position: relative`, `position: absolute`, `position: fixed` to `<Node.svelte>` (data-type="node") in edit mode. Only `position: static` is permitted to allow css anchor positioning queries to resolve correctly.
 - Never use an `<a>` tag inside a `contenteditable="true"` element, as it will cause unexpected behavior. Make it a `<div>` while editing, and an `<a>` in read-only mode (when `svedit.editable` is `false` ).
-- Avoid CSS selectors like `:last-child` or `:first-child` on nodes (e.g., `[data-type="node"]:last-child`) because `<NodeGap>` elements are inserted in edit mode and will become the actual first or last child. This can cause unexpected layout shifts (e.g., if you have `.paragraph-node:last-child { margin-bottom: 10px }`, the margin won't apply as expected).
-- Avoid adding css `margin` to nodes inside node arrays when using flex or grid layouts. Use `gap` on the container instead. This ensure `NodeGap` and `NodeGapMarkers` render consistently. If you need to add a margin, add it to child element of Node.
+- Avoid adding css `margin` to nodes inside node arrays when using flex or grid layouts. Use `gap` on the container instead. This ensures `NodeGap` and `NodeGapMarkers` render consistently. If you need to add a margin, add it to a child element of Node.
+
+### Consistent DOM structure across modes
+
+`<NodeArrayProperty>` always renders a `.node-gap` element before each node, regardless of whether the editor is in edit or read-only mode. In edit mode, this is the full `<NodeGap>` component with anchor positioning, hit areas, and caret support. In read-only mode, it's a plain `<div class="node-gap"></div>`.
+
+This means the DOM structure is identical in both modes:
+
+```html
+<!-- Same structure in edit and read-only mode -->
+<div data-type="node_array">
+  <div class="node-gap">...</div><!-- First element is a node-gap -->
+  <div data-type="node"><!-- first node --></div>
+  <div class="node-gap">...</div>
+  <div data-type="node"><!-- second node --></div>
+  ...
+  <div class="node-gap">...</div><!-- Last element is a node-gap -->
+</div>
+```
+
+The `.node-gap` element uses `display: contents`, so it has no layout impact — it doesn't generate a box, doesn't affect flex/grid item counts, and doesn't consume space. It only serves as a DOM placeholder for consistent selector targeting.
+
+Because `.node-gap` is always the actual first and last child, do not use `:first-child` or `:last-child` on nodes — they will never match. Use `:nth-child(1 of [data-type="node"])` and `:nth-last-child(1 of [data-type="node"])` instead.
+
+CSS sibling selectors that target node adjacency only need one form:
+
+```css
+/* Works in both edit and read-only mode */
+.node-text + .node-gap + .node-media {
+  margin-block-start: 1rem;
+}
+```
+
+For `:nth-child` selectors, use the `of <selector>` syntax to skip over `.node-gap` elements and count only nodes:
+
+```css
+/* Even nodes */
+.list-node-array > :nth-child(2n of [data-type="node"]) {
+  background: var(--zebra-stripe);
+}
+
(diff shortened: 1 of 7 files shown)


johannesmutter:
This one needs to merge first before the other PRs @michael to avoid merge conflicts

michael:
Will need a bit of time for the reviews (afk for most of the day).

One issue I spotted (present in main but related to this PR) is that for empty-node-array/placeholder we now have additional cursor positions.

Reproduce:
- set the cursor at the end of story.description
- press arrow-right
- press arrow right again
- (cursor is one time inside empty-node-placeholder, and another time within node-gap

Related code:

```svelte
{#if svedit.editable}
	<div
		class="empty-node-placeholder"
		data-path={[...path, 0].join('.')}
		data-type="node"
		style="anchor-name: --{[...path, 0].join(
			'-'
		)}; min-height: 40px; min-width: 24px;"
	></div>
	<!-- Sibling (not child) of .empty-node-placeholder so its .svedit-selectable
	     resolves anchor positioning against the shared containing block,
	     not the placeholder which inherits .node positioning styles. -->
{/if}
<NodeGap
	array_path={path}
	offset={0}
	count={0}
	empty
	positioned={svedit.editable ? (svedit.should_position_gap?.(path_str, 0, 0) ?? true) : false}
/>
```

The reason is that an empty div creates an addressable cursor position. Also any whitespace between the elements can create cursor positions. I quickly tried setting contenteditable="false" on the empty-node-placeholder, which fixes it for arrow-down navigation, but creates more additional cursor positions (which select the whole node) for when navigating with arrow-right.

For reference, in v0.6.4 this was working correctly. See: https://github.com/michael/svedit/blob/v0.6.4/src/lib/NodeArrayProperty.svelte#L23-L38

Maybe you need to revert to old DOM structure for that one edge case.

johannesmutter:
After a few iterations, here's the path we walked. Posting this so the final shape makes sense in context.

### The bug

For an empty `node_array` (e.g. an empty `buttons`), the placeholder div + the sibling `NodeGap`'s `.svedit-selectable` each created an arrow-key cursor stop in the same logical position. Pressing right twice was needed to traverse the empty area. 

### What we tried

1. **Nest `NodeGap` inside `.empty-node-placeholder` (v0.6.4 layout).** Killed the wide click area — Story.svelte's `position: absolute` on the placeholder makes it the containing block, so `.svedit-selectable`'s anchor-based expansion got clamped to the small placeholder box.
2. **`contenteditable="false"` on the placeholder.** Atomic-element behaviour created before/after stops in the parent that selected the whole placeholder — worse than the original. (Matches what you'd already noticed.)
3. **`inert` on the placeholder.** Same atomic-stop problem.
4. **Drop the `<br>` from `.svedit-selectable` when empty.** Empty editable `<div>` is still a caret stop in Chromium, so still two stops.
5. **`contenteditable="false"` on `.svedit-selectable` + `onpointerdown` redirect + `__render_node_selection` special case.** Worked end-to-end but ~49 lines across three files, leaked the empty-placeholder concern into `Svedit.svelte`'s render path, and the `contenteditable` flip stranded the caret on add → delete cycles (atomic element traps the caret when the array becomes empty mid-edit).
6. **Just `<br>` in placeholder + skip `+1` for empty arrays.** 10 lines, killed the validation error, but two DOM stops still existed — visually they overlap so the cursor doesn't appear to move, but the user still pressed right twice to leave the area.
7. **Imperative `data-cursor-here` mirror in `onselectionchange`.** 21 lines, worked, but leaky — Svedit had to know NodeArrayProperty's class names, the handler ran on every keystroke (even nowhere near a placeholder), and we'd be coupling two files via DOM strings.
8. **Finally `.empty-node-placeholder:focus + .node-gap { display: none }`.** Right shape, but `:focus` doesn't match contenteditable child divs — the canvas owns focus, so the rule never fires.

### Where we landed (current PR state)

A reactive `is_focused` derived state, mirroring the pattern already in `AnnotatedTextProperty.svelte`. The class lives on the `node_array` root so it's a general "this array holds the caret" signal — reusable for other selectors later. Useful also e.g. if you want to style active node array in app land.

michael:
Not yet fully happy with this solution, as the root of the issue is too many selectable cursor positions are on the canvas for empty node arrays, which are now worked around by hiding some during the is_focused state.

Haven't yet been able to debug what causes the additional cursor positions in this particular case, but happy to take this over and figure it out. 

(Always a bit tricky to narrow down selectables, but it's worth it that one can always be sure the cursor can land in exactly one place. For selection mapping in particular.)

johannesmutter:
well we can just remove the br from inside empty-node-placeholder and wrap empty-node-placeholder around the nodegap.

johannesmutter:
Simplified and fixed. Please take a look @michael

johannesmutter:
I kept the `.focused` class since its very handy to know in css which element in the editor has currently the focus (same pattern exists on `AnnotatedTextProperty.svelte`

johannesmutter:
Just to recap what was the issue: empty-node-placeholder had a br, as well as the NodeGap. So two traps for the cursor when the node array was empty

michael:
Yes, now there's just one cursor position left! 👍

I suspect that now this change in the selection mapping (Svedit.svelte) is also not necessary anymore. But trying to recall how that "cursor is in empty-node-array-node-gap" case was detected and handled before.

johannesmutter:
> I suspect that now this change in the selection mapping (Svedit.svelte) is also not necessary anymore. But trying to recall how that "cursor is in empty-node-array-node-gap" case was detected and handled before.

The guard is technically correct (bumping to offset 1 in an empty array would be out of bounds), but the case is never reached. The early return at line 814 catches any collapsed selection inside a gap, which is the only way the caret enters an empty array.

johannesmutter:
Now reverted. Ready to merge?

michael:
> The guard is technically correct (bumping to offset 1 in an empty array would be out of bounds), but the case is never reached. The early return at line 814 catches any collapsed selection inside a gap, which is the only way the caret enters an empty array.

Yeah, just for explanation: We don't want to guard things that are considered invalid state, that's where we actually want to see the exceptions. The out of bound exception was actually making me aware of the problem. So given the exceptions are implemented correctly the selection error means something is wrong with the state of the document/selection (which should be addressed - like in our case get rid of the accidental cursor position), not to add a handler to degrade gracefully - a recipe for never detecting invalid state.

Similar here: https://github.com/michael/svedit/issues/262. We want to fail early because Svedit currently has an invariant "annotations must not overlap", we just didn't enforce them yet. So what we want is to **fail early** (important principle), not to let the system slip into invalid state, and make the implementation robust against it. Will create an issue for that.

Merging this now...

johannesmutter:
> Similar here: #262. We want to fail early because Svedit currently has an invariant "annotations must not overlap", we just didn't enforce them yet. So what we want is to **fail early** (important principle), not to let the system slip into invalid state, and make the implementation robust against it. Will create an issue for that.

Curious how do you think can we fail early when it concerns a database operation? When data is corrupted? Should there be rather a invisible cleanup happening?

michael:
basically what i mean is:

- when data is corrupted: throw an error
- so you can fix the data (e.g. in the database)
- then run the app reliably


but also:
- if use the svedit apis to create data using svedit.session, it should fail e.g. on a create or set operation when invalid data is assinged.

The goal is to not let invalid data slip through (when ever possible, of course there are infinite ways to break a system, and you might never be able to catch all)

PR #266, opened 2026-05-05
Fix selection mapping regressions and remove `text-box: trim-both` for now

michael:


michael:
This fixes a range of bugs related to selection mapping and refocusing the canvas/window. It should be much better across all browsers now.

It also fixes a bug where Firefox editing was breaking during dead-key character composition.

I figured that `text-box: trim-both` is not yet supported by Firefox, which means it's too early to make it a default, as you'd have to use different spacings/paddings for Firefox so pages render the same across browsers. Will switch to `text-box: trim-both cap alphabetic` as soon as all browsers support it. Meanwhile apps can activate text-box: trim-both in userland.

PR #267, opened 2026-05-06
Make Transaction.set ref-count-aware when cleaning up removed nodes

johannesmutter:
Transaction.set previously called this.delete() unconditionally for any node ID that was present in the property's previous value but not in the new value. This is wrong whenever the removed ID is still reachable from elsewhere in the document — most notably when a single transaction creates a wrapper that references the children before setting the parent array to drop them (the "wrap-in-group" pattern used by editors built on Svedit, e.g. interface-lineage's WrapInGroupCommand). The unconditional delete force-destroys the children and leaves the wrapper dangling.

Replace the loop with a call to the existing _cascade_delete_unreferenced_nodes helper, which only deletes nodes whose reference count has dropped to zero. Document the asymmetry on delete() — it stays a force-delete on purpose, because annotate_text and other callers need to remove a node while a stale reference still exists in the property they are about to update.

Tests: add two cases to the Deletion scenarios suite covering the forward wrap-style move and its undo.

### DIFF
--- a/src/lib/Transaction.svelte.js
+++ b/src/lib/Transaction.svelte.js
@@ -223,9 +223,15 @@ export default class Transaction {
 		this.inverse_ops.push(['set', normalized_path, previous_value]);
 		this._apply_op(op);
 
-		for (const removed_node_id of removed_node_ids) {
-			// NOTE: This implicitly deletes childnodes as well, given that they are no longer referenced.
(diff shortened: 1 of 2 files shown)


michael:
Good catch. Makes sense. Will merge after #255 is merged.

Issue #268, opened 2026-05-06
Anchor positioning doesn't work as expected when anchor is inside scrollable container


michael:
There's one issue (also present in main, but not in v.0.9.0) where anchor positioning at the overlay layer doesn't work as expected for scrollable containers.

It seems the popover's target is now the whole container, not the focused nav item node anymore. Scrolling horizontally doesn't move the popover.

https://github.com/user-attachments/assets/ae23b4b8-1d5b-4fce-b850-278b62344f6e

In 0.9.0 this part works (but there node gaps do not reposition as you scroll - only after scroll stopped):

https://github.com/user-attachments/assets/1508b957-7b0d-49d7-a191-26ce7ece99d4

_Originally posted by @michael in https://github.com/michael/svedit/issues/255#issuecomment-4387548159_

michael:
This is actually working fine, I just had my anchor names wrong.

Might finish #270 at some point to bring some floating popovers into the demo, for editing link and image urls, but it's not a blocker for 0.10.

PR #270, opened 2026-05-06
[WIP] Popovers and scrollable containers

michael:


michael:
This is unpolished, and unfinished, but a starting point to have a demo/playground to improve popover positioning when overflowing containers are involved (see #268)

This might only partly be a svedit concern. The second part could be a recommendation/example how to implement popovers in your app, that will work in combination with scrollable containers.

michael:
NOTE: the current clamp to screen edges logic doesn't respect the scroll position in the overflow container, which is something we should address.

interestingly the positioning (centered, below the anchor) generally works in this example in Svedit, while in editable website when using svedit#main the popover is rendered on top of the anchor. So i still don't understand why with svedit v0.9.0 it worked, but not anymore. so a bit confused why it works here in svedit as i'd expect (leaving aside the clamp to screen edge problem)

michael:
> interestingly the positioning (centered, below the anchor) generally works in this example in Svedit, while in editable website when using svedit#main the popover is rendered on top of the anchor. So i still don't understand why with svedit v0.9.0 it worked, but not anymore. so a bit confused why it works here in svedit as i'd expect (leaving aside the clamp to screen edge problem)

ok my bad. the positioning in editable website broke because of the changed anchor name format in svedit#main and my code was relying on the old format

michael:
Discarding this for now, as it's too much app-infra for a demo. And the popover placement is an app-concern as well, which Svedit doesn't prescribe.

Issue #276, opened 2026-05-09
Remaining issues with node gaps in scrollable containers


michael:
When you enter a couple of buttons so they overflow...

- [ ] they are not scrolled into view
- [ ] the last node gap is missing (it comes back, once you navigate with the keyboard later, so must be some visibility culling/checking issue)

You can reproduce on https://svedit.dev.

https://github.com/user-attachments/assets/55b64162-89ca-49b1-a6d1-0e25f36dcc33

michael:
Also noticed: It seems that sometimes the scrollable container can get a focus, and then no cursor nav, no keyboard shortcuts (ENTER) are working. I guess we might want to find a way to make scrollable containers not focusable (currently just hiding focus outline)

johannesmutter:
Two new PRs addresses all these issues: https://github.com/michael/svedit/pull/278
https://github.com/michael/svedit/pull/279

johannesmutter:
https://github.com/user-attachments/assets/17803424-d8c7-4041-8fda-9e2705c47a0f

michael:
Fixed via: https://github.com/michael/svedit/pull/278

PR #278, opened 2026-05-11
Node gap overflow fixes

johannesmutter:
Addresses these issues here: https://github.com/michael/svedit/issues/276


johannesmutter:
Performance tests show overall same, with a bit faster performance at mutations.

BEFORE CHANGES:

| 50    | 68        | 673    | 21   | 31     | 60         | 56         | 60         | 57         | 32           | 20           |
| 100   | 74        | 1353   | 19   | 30     | 58         | 29         | 60         | 56         | 29           | 20           |
| 200   | 134       | 2715   | 18   | 32     | 41         | 20         | 53         | 29         | 22           | 15           |
| 500   | 305       | 6825   | 18   | 52     | 19         | 12         | 17         | 9          | 8            | 7            |
| 1000  | 719       | 13681  | 17   | 113    | 7          | 3          | 4          | 2          | 3            | 3            |
| 2000  | 2404      | 27374  | 17   | 171    | 1          | 1          | 1          | 1          | 0            | 0            |


AFTER CHANGES:

| 50    | 67        | 670    | 18   | 38     | 60         | 57         | 60         | 56         | 55           | 30           |
| 100   | 80        | 1350   | 16   | 53     | 58         | 30         | 60         | 56         | 39           | 20           |
| 200   | 132       | 2713   | 16   | 61     | 43         | 29         | 60         | 56         | 34           | 30           |
| 500   | 315       | 6822   | 15   | 59     | 17         | 12         | 20         | 12         | 14           | 12           |
| 1000  | 776       | 13678  | 14   | 117    | 7          | 4          | 4          | 3          | 4            | 3            |
| 2000  | 2518      | 27371  | 14   | 183    | 2          | 1          | 1          | 1          | 1            | 0            |

---

AFTER CHANGES WITHOUT CULLING ENABLED:

| 50    | 187        | 736    | 84   | 32     | 60         | 56         | 34         | 20         | 15           | 12           |
| 100   | 443        | 1507   | 173   | 24     | 49         | 20         | 20         | 12         | 8           | 7           |
| 200   | 1237       | 3044   | 347   | 32     | 39         | 12         | 10         | 5         | 4           | 3           |
| 500   | 6587       | 7679   | 872   | 54     | 22         | 4         | 3         | 2         | 1           | 1           |
| 1000  | 25823       | 15412  | 1748   | 94    | 28          | 2          | 1          | 1          | 9            | 0            |

2000 nodes can't render without culling.

michael:
Reverted the change regarding selecting button.label after insertion. Only temporarily so the root issues can be fixed first, which are still present.

- [x] setting a node caret should scroll the affected node gap into view, when in an overflowing container
  - NOTE: there may be other situations where this is needed, e.g. when a node cursor that is not in view gets restored after after an undo; or for an horizontally scrollable image gallery to add more gallery items fast; 
- [x] last node gap disappears in some situations (see repro)

Reproduce:
- Add 5 buttons with ENTER ENTER ENTER so they overflow (node cursor not scrolled into view)
- now remove a couple of them so the buttons no longer overflow (now the last node gap is missing)

https://github.com/user-attachments/assets/5628eabd-29e6-4117-8f1d-5bdfb6e055f5

The issue with scrolling node gaps into view is not as pressing - i can look into myself, the one with the disappearing node gap i could need help with, as i don't yet fully grasp how the node gap rendering works in detail.

The 'last node gap disappears' problem is also present when visibility culling is turned off. Also i noticed that when `visibility_culling: false` that we get a horizontal scrollbar for the body when you create a lot of buttons.

johannesmutter:
Fixed

michael:
I think the 'last node gap disappears' looks good now.

But the scroll to node selection fix has regressions. When making a node selection and the node is already fully in the viewport, a body scrollTo is triggered all the time.

Reproduce:
- add new button with ENTER (triggers unexpected body scroll)
- or select with the mouse starting from story.title release mouse over story.description (selection changes to node selection and body scroll is activated, dragging the viewport away while you're selecting)

This line causes the regression during mouse dragging as Svedit must not rerender node selections (any selection really), unless selection derived from dom !== current model_selection -> only then it needs a refresh, otherwise creates a battle-loop with the current selection that is still being dragged
https://github.com/michael/svedit/pull/278/changes/1d279005cfd2beb59e125398322f4101942a2cea#diff-fd1f70e683b5b47eb06507e6fd0743f688787950b09e6a687beb76efaf2f4a8bR755

When removed, the unexpected body scroll after ENTER is still there though. I think in __render_node_selection the logic could be as simple as: scroll to the .node-gap node for collapsed node cursors or for single+multi-node selections scroll into view the beginning of the first node or end of the last node (depending on value of is_backward) of a node selection (should work for single node selections too)

johannesmutter:
| 50    | 85        | 670    | 18   | 42     | 60         | 56         | 60         | 56         | 50           | 20           |
| 100   | 118       | 1350   | 16   | 32     | 53         | 20         | 60         | 57         | 43           | 29           |
| 200   | 128       | 2713   | 16   | 36     | 36         | 15         | 56         | 30         | 27           | 20           |
| 500   | 388       | 6822   | 15   | 51     | 18         | 12         | 19         | 12         | 13           | 10           |
| 1000  | 814       | 13678  | 14   | 104    | 6          | 3          | 4          | 2          | 3            | 2            |
| 2000  | 2970      | 27371  | 14   | 181    | 1          | 1          | 1          | 1          | 1            | 1            |

michael:
Skipping based on selection_source_is_dom (instead of DOM selection compare) might be a great idea, and seems to fix some other timing-based issues as well (e.g. sometimes composition of á with deadkeys leads to the cursor being placed at 0 instead of 1). Doing more extensive testing to make sure this doesn't regress elswhere, but looking good so far.

Regarding the focus blur fix when a node array container gets the focus: I feel like this focus should not be possible in the first place, because it creates some weird stalled selection that doesn't map to anything. Selection mapping would be skipped in that case and model selection keeps its previous state, which violates the DOMSelection===ModelSelection invariant. But is there actually a way to make sure that in edit mode a node array container can never get the focus? I guess tabindex=-1 is not enough here to prevent the container from getting focused, or is it?

johannesmutter:
Both points addressed.
- **`selection_source_is_dom`**: kept. Working as you observed.
- **Node-array focus**: you were right — Svedit was calling `node_array_el.focus()` on itself. Removed. Same pass also dropped `el.focus()` in `__render_text_selection` (the `// needed?` one) and the now-orphaned `relatedTarget` blur guard.
- Replaced `focus_retention` (assertions too loose) with `keymap_continuity`: dispatches real `Enter` after an insert, asserts the command fires.
- Side cleanup: test runner is now headless playwright. `npm test` for normal use, `npm run test:visual` to watch in your browser. It's less annoying if the tests run in the background and don't open a new tab in different browser window on each run.

michael:
Thank you very much! I'll take a closer look as soon as I got some sleep, by brain can't be trusted today.

I think the fixes in Svedit are solid and much improve overall stability, not just making the issues go away!

As for the changes in node_visibility.svelte.js, maybe you can give me a quick recap, what changed and why, in a few sentences, that will help me bevore i dive into the code/commit messages etc? Are the changes soley addressing the issue of disappearing node gaps, or was there other problems too you addressed?

michael:
Oh and can you maybe revert the playwright commit and make it a separate PR? Can't run the tests anymore. Happy to give playwright another chance, but i remember i had all kinds of weird issues with it back then, and then went with browser mode only, as it's more lightweight and i could be sure there's no headless browser flakiness etc. but maybe things improved since then.

johannesmutter:
> As for the changes in node_visibility.svelte.js, maybe you can give me a quick recap, what changed and why, in a few sentences, that will help me bevore i dive into the code/commit messages etc?

Mostly addresses the disappearing-gap issue, with two small side improvements.

The core fix: replaced `clip_map` (per-node, IO-driven) with `edge_map` (per-array, scroll-driven). `clip_map` was stale on horizontal scroll inside a nested overflow container because IntersectionObserver doesn't refire when the intersection ratio doesn't cross a threshold. `edge_map` is computed from the array's own `scrollLeft`/`scrollWidth`/`clientWidth`: six property reads, updated by a document-level scroll listener that early-outs in O(1) for non-array scrolls.

Side improvements :
- `flush_view_sync` split into read-pass + write-pass to avoid layout thrash when consumer CSS attaches layout-affecting rules to `.in-view` / `.fully-in-view`
- Added a `visibility_culling` config flag for debugging (set to false → bypass IO, all nodes treated as near — useful to isolate culling bugs from anchor-positioning bugs)
- Cleanup: removed `transfer_path` + the MO attribute branch (defensive code for a hypothetical re-keying that never actually fires in production); dropped IO threshold from `[0, 1]` to `[0]` since `#process_near` only reads `entry.isIntersecting`

michael:
When trying this branch using SvelteKit's alias against editable-website (branch [svedit-0-10-0](https://github.com/michael/editable-website/pull/109/changes)), I do still have missing last node gaps in some node arrays, e.g. here:

image

That exact problem is not there when I use the svedit#main branch.

Is there something that has changed API-wise that I need to update in the app maybe? I can't currently reproduce that problem directly in Svedit.

michael:
Maybe the reason is that the scroll metrics are now always used for showing/hiding gap markers. But when the container has normal visible overflow, like this LinkCollection in my case, they should just always be shown, not clipped.

johannesmutter:
Can you check if there's some specific tailwind styles at rule here? e.g. overflow, position rules, etc.?

>  I do still have missing last node gaps in some node arrays, e.g. here:

is that the only one, or are there others too?

michael:
Really no special positioning rules here, just a grid layout.

```svelte
<NodeArrayProperty
	class="grid grid-cols-1 md:grid-cols-2 xl:grid-cols-3 gap-16 md:gap-10 lg:gap-14 [--row:1]"
	path={[...path, 'link_collection_items']}
/>
```

And the link collection item:

```svelte
<Node class="link-collection-item group" {path}>
	<svelte:element
		this={render_as_link ? 'a' : 'div'}
		href={render_as_link ? node.href : undefined}
		target={render_as_link ? node.target : undefined}
		class="block outline-1 outline-transparent focus-visible:outline-1 focus-visible:outline-(--svedit-editing-stroke) focus-visible:outline-offset-1"
		use:reveal
	>
		<div class="overflow-hidden" style:aspect-ratio="4/3" style:border-radius="var(--image-border-radius)">
			<MediaProperty path={[...path, 'media']} />
		</div>
		<div class="pt-4">
			<AnnotatedTextProperty class="text-xs md:text-sm uppercase tracking-widest text-(--foreground) opacity-60 mb-2" path={[...path, 'preline']} placeholder="Preline" />
			<div class="pt-1">
				<AnnotatedTextProperty class="{!svedit.editable ? 'title-underline' : ''} inline font-serif text-(--foreground) text-2xl lg:text-3xl text-balance" path={[...path, 'title']} placeholder="Title" />
			</div>
			<AnnotatedTextProperty class="text-balance pt-2" path={[...path, 'description']} placeholder="Description" />
		</div>
	</svelte:element>
</Node>
```

This change fixes it for me, not sure if that's the best way to address it though.

image

johannesmutter:
can you check against previous commits of this branch if one worked?

johannesmutter:
Also do you have URL parameter like ?edit=true to always test in editing mode?

michael:
> can you check against previous commits of this branch if one worked?

The first commit of this branch already breaks it https://github.com/michael/svedit/pull/278/changes/821a28dc9d1ba094a7d45af006fcf41008a7f652  (with main the last node gap is displayed in this situation.)



FYI: doesn't just happen in one place, all grid-containers are affected, for the gallery the last node gap is also missing.

> Also do you have URL parameter like ?edit=true to always test in editing mode?

No, I start in read mode then turn edit mode on.

johannesmutter:
ok so I tried various solutions, your fix is the easiest.

michael:
Ok, I pushed the fix. Will do some more manual testing from within EW. If nothing pops up, this will be good to merge then?

michael:
Ah sorry, I see you were faster, no commit from me then.

PR #279, opened 2026-05-11
Break text on block

johannesmutter:
New behaviour: 
Determine which text property holds the content to split. For a 'text' kind node it's always 'content'. For a 'block' kind node we only proceed when this is the node's **SOLE** text-shaped property (annotated_text / text). Multi-text-field blocks (story, hero) stay caret-at-gap. So a button in a buttons array behaves the same as a paragraph in the body array:

The dominant intention of a user is to create a button and then change the label / add a link.

Previously it would focus the node gap afterwards, so an extra arrow key back stroke was necessary to focus the button content. Also previously pessing enter inside a button didn't do anything. now it also create a new button node afterwards. consistent behaviour with paragraph.

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -1534,7 +1534,7 @@ ${fallback_html}`;
 
 	@media not (pointer: coarse) {
 		@supports (anchor-name: --test) {
-			:where(.svedit-canvas.hide-selection) :global(::selection) {
+			.svedit-canvas.hide-selection :global(::selection) {
 				background: transparent;
 			}
 		}
--- a/src/lib/transforms.svelte.js
+++ b/src/lib/transforms.svelte.js
@@ -20,16 +20,39 @@ export function break_text_node(tr) {
 	// First we need to ensure we have a text selection
 	if (selection.type !== 'text') return false;
 
-	// Next, we need to determine if the enclosing node is a pure text node (e.g. paragraph),
-	// which is wrapped inside a node_array (e.g. page.body)
+	// Next, we need to determine if the enclosing node is a pure text node
+	// (e.g. paragraph) OR a single-text-field block node (e.g. button),
+	// which is wrapped inside a node_array.
 
-	// Owner of the text property (e.g. paragraph)
+	// Owner of the text property (e.g. paragraph or button)
 	const node = tr.get(selection.path.slice(0, -1));
-	if (tr.kind(node) !== 'text') return false;
+	const kind = tr.kind(node);
+	const text_prop = selection.path.at(-1);
+
+	// Determine which text property holds the content to split. For a
+	// 'text' kind node it's always 'content'. For a 'block' kind node
+	// we only proceed when this is the node's SOLE text-shaped property
+	// (annotated_text / text). Multi-text-field blocks (story, hero)
+	// stay caret-at-gap — the user picks which field by gap selection,
+	// per the project rule.
+	if (kind === 'text') {
+		// existing behaviour
+	} else if (kind === 'block') {
+		const node_schema = tr.schema[node.type];
+		if (!node_schema?.properties) return false;
+		const text_props = Object.entries(node_schema.properties).filter(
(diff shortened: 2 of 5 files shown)


johannesmutter:
@michael This branch depends on changes in node-gap-overflow-fixes. That's why it's setup to merge into that branch and not main. So review is easier for you.

michael:
I think it would be better to treat story.buttons as kind: 'text' to inherit the text split/join behavior, rather than implementing text/split behavior for kind: 'block' where there is just one text property.

Currently "text nodes" assume that there's one annotated_text property called 'content'. Furthermore they assume that there are no additional properties other than layout.

But we could generalize this and say kind: 'text' requires your block to have just one annnotated_text property, which you can name as you please (e.g. button.label suits better than button.content). "text blocks" could then also carry any other non-text properties (such as button.href).

However, there are implications:

If "text nodes" carry other information than the text the split and join operations might lead to surprising results for users. Let's say you have 2 buttons, first linking to foo.com, second to bar.com. Now you put the cursor at the first position and hit backspace. The node gets joined with the previous one, with a combined label and foo.com as href. Now let's say you press ENTER right after: now you visually are where you were before, with a subtle difference: The second button will now hold foo.com for the href (because on split, we can just copy properties over, the bar.com really is lost at the join operation) which the user might overlook and be surprised.

Hmm... honestly, I think that for story.buttons it's better to behave like full blocks with no join/split text behavior. Because that makes it obvious that each button is a "unit of information" (a block), not just text that can be joined/splitted. What we are trying here is make a structural block feel like just text, but because of the extra url property (there might be even more in similar use-cases) it's not just text.

Wdyt?

> Previously it would focus the node gap afterwards, so an extra arrow key back stroke was necessary to focus the button content.

Putting the cursor into the label, instead of putting a node caret after the inserted button would make sense though. This is already up for the app to decide in the inserter logic. Should we change it in Svedit? The good thing about the current behavior is that it's easier to test the overflow thing with ENTER ENTER ENTER, so maybe keep it for Svedit? But for apps, yes focusing the label makes more sense. I will make that the behavior for nav items in EW.

johannesmutter:
The difference between text nodes and block notes is a bit ambiguous/ buried in the internals of svedit. If there's really a difference in expected behaviour (which I think there is), than this should be made more clear, and using it the "wrong" way should be either prohibited or warned in the console. 

> Now you put the cursor at the first position and hit backspace. The node gets joined with the previous one, with a combined label and foo.com as href. Now let's say you press ENTER right after: now you visually are where you were before, with a subtle difference: The second button will now hold foo.com for the href (because on split, we can just copy properties over, the bar.com really is lost at the join operation) which the user might overlook and be surprised.

I wouldn't find that very surprising, as you get visual feedback along the way: the buttons became one. The last one joined the first one. The first one wins. The second looses its additional information. It's e.g. the same with lets say a block background color. First one had red, second blue. Now backspace, the texts are joined, and there's just a red. Split again and the expecation would be two red, not one blue one red. So it really depends on the type of property. And if something depends, it means its rather a sign to making opting in & out into a behaviour more easy to developers: allow join on this node: yes/ no?

But I haven't really looked at a variety of such examples to draw a conclusion on wether the behaviour should be baked into the node type, or wether it should be something you add to a node type, with good defaults.

> Putting the cursor into the label, instead of putting a node caret after the inserted button

I think we should do this. The principles: respect the users intend. Reduce the users muscle effort.

> is that it's easier to test the overflow thing with ENTER ENTER ENTER

The svedit landing page should not be optimised for easy testing, but rather best practises to copy / inspire app developers. I think we in general need a more comprehensive test page, that covers really ALL layout relevant css properties, in combination with all svedit primitives, a place to go for things writing unit tests for is too much effort.

johannesmutter:
fixed the merge conflics

michael:
Convinced. Added the discussed behavior for story.buttons, but using the existing kind: 'text' mechanism. In many cases this will feel the most natural for single-text field nodes. But your app can always decide otherwise and make it a "hard block" instead.

See: https://github.com/michael/svedit/pull/281

PR #284, opened 2026-05-20
Fix empty node array gap not filling its placeholder

johannesmutter:
An empty node array's gap (`.svedit-selectable`) had no usable hit area.

## Root cause

The empty gap's selectable was sized by `min(anchor(--placeholder), anchor(--container))` and gated on the `.positioned` class. Two compounding failures:

1. **It never got `.positioned` on initial render.** `.positioned` is toggled by `sync_gap_class()`, reached via `#process_near` → `sync_gaps_around_node()` — which only inspects a node's *siblings* for a `.node-gap`. An empty array's gap is nested *inside* the `.empty-node-placeholder` as a *child*, so it was never synced; its selectable stayed collapsed at `0 × 0`.

2. **Even positioned, the anchor math collapsed it.** Where the placeholder is absolutely positioned (e.g. the Story `buttons` array), `anchor(--container …)` resolved only the start-side edges — `right`/`bottom` fell through to `auto`, and the selectable shrink-wrapped to its `<br>`: a zero-width hit area.

## Fix

An empty array's gap doesn't need anchor positioning. Anchor positioning places a gap in the whitespace *between sibling nodes*; an empty array has none. Its gap just has to fill one box — the placeholder.

- **`NodeArrayProperty.svelte`** — `.empty-node-placeholder` is now `position: relative`, making it the containing block for its gap's selectable.
- **`NodeGap.svelte`** — the empty-gap selectable fills the placeholder with `inset: 0`. No `anchor()`, so no `.positioned` gating and no anchor-resolution cost; the rule applies unconditionally.

The hit area is now exactly the placeholder — which the app already sizes (in flow, out of flow, or zero) — and `node_visibility` no longer touches the empty gap at all.

## Test plan

- [x] Empty node array on screen: its `.svedit-selectable` fills the placeholder (real size, `pointer-events: auto`).
- [x] A pointer gesture inside an empty array reaches the gap.
- [x] Works with an absolutely-positioned placeholder (Story `buttons`).
- [x] Non-empty arrays: gap positioning unchanged.

### DIFF
--- a/src/lib/NodeArrayProperty.svelte
+++ b/src/lib/NodeArrayProperty.svelte
@@ -96,11 +96,14 @@
 </svelte:element>
 
 <style>
-/* If you override this to position: absolute (e.g. so an empty array
-   doesn't occupy flow space), set position: relative on the parent
-   node-array container so left/right or inset resolve against it. */
+/* position: relative makes this the containing block for the gap's
+   .svedit-selectable, which fills it via inset:0. You may override to
+   position: absolute (e.g. so an empty array doesn't occupy flow space)
+   — still a containing block; then also set position: relative on the
+   parent node-array container so this placeholder's inset:0 resolves
+   against it. */
 	:where(.empty-node-placeholder) {
-		position: static;
+		position: relative;
 		inset: 0;
 		min-height: 40px;
 		min-width: 24px;
--- a/src/lib/NodeGap.svelte
+++ b/src/lib/NodeGap.svelte
@@ -358,12 +358,21 @@
 		min-width: calc(var(--_eg) * var(--_R));
 	}
 
-	/* Empty array: spans the outermost edges of placeholder and container. */
-	:global(.node-gap.positioned.gap-before.empty .svedit-selectable) {
-		top: min(var(--_s-t), var(--_c-t));
-		bottom: min(var(--_s-b), var(--_c-b));
-		left: min(var(--_s-l), var(--_c-l));
-		right: min(var(--_s-r), var(--_c-r));
+	/* Empty array: the gap fills its placeholder, which is this
+	   selectable's containing block (NodeArrayProperty sets it
+	   position: relative). inset:0 fills it with no anchor() — so no
+	   .positioned gating and no anchor cost. The width:auto, height:auto
+	   and pointer-events:auto override the :not(.positioned) 0×0
+	   collapse; position-visibility:always overrides the .positioned
+	   rule's anchors-visible, so a transient positioned toggle can't
+	   hide the gap. */
+	:global(.node-gap.gap-before.empty .svedit-selectable) {
+		position: absolute;
+		inset: 0;
+		width: auto;
+		height: auto;
+		pointer-events: auto;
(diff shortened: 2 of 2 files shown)


michael:
Just so I understand correctly, the issue here is related to gap positioning only, right? Because your initial code suggested that it affects selection mapping, but there's no issue regarding that right?


Diving into this topic, while testing with EW I also realized we may want to improve the situation around empty node placeholder placement. Like should it affect the layout (view/edit mode differs) or should it never affect the layout.

Created a separate issue for that: https://github.com/michael/svedit/issues/285 (to be discussed and implemented outside of this PR, not to mix things up)

johannesmutter:
previous title and message was false direction/ different symptom, not the root cause of the issue I experienced.
I added a right click event that required a click on the empty node array, but didn't resolve because of this issue.

michael:
Can you give me steps to reproduce at svedit.dev to see the issue this fixes? Once i can reproduce i will merge, unless i see regressions.

johannesmutter:
In svedit.dev inspect the svedit-selectable inside a .empty-node-placeholder. it has height and width zero. as it doesn't get a positioned class. Therefore it doesn't have any clickable area for custom events.

johannesmutter:
Screenshot 2026-05-20 002000 Svedit - A tiny library for building editable websites in Svelte@2x

michael:
Got it, thanks. In this branch the .svedit-selectable has a height now, but width is still 0. Is that alright?

image

johannesmutter:
I'm just investigating this issue. It's I think because of the absolute position the node itself has no size ... the svedit-selectable should maybe rather query the empty-node-placeholder in this case for its anchor positioning

michael:
Yes, for static empty placeholder node gaps the dimensions are correct: 

image

johannesmutter:
ok so this entire PR is looking at the wrong problem I think …

johannesmutter:
Rewrote the description — the original root-cause writeup was wrong. It pinned this on MutationObserver / hydration timing; that isn't it.

Confirmed in the browser: the empty placeholder *is* observed and `#process_near → sync_gaps_around_node` does run for it (it picks up`.in-view` / `.seen` / `.fully-in-view`). The gap is missed purely because `sync_gaps_around_node` inspects a node's *siblings*, and an empty array's gap is nested inside the placeholder as a *child*. I forgot why it is nested, but there was a very important reason for it.

johannesmutter:
Reworked — this is now a CSS fix, not a `node_visibility` one.

The earlier approach taught `node_visibility` to position the empty gap. That worked, but it left the gap's *size* still computed by the `anchor()` math — which collapsed to a zero-width hit area where the placeholder is absolutely positioned (the Story `buttons` case).

An empty array's gap doesn't need anchor positioning at all — it just has to fill its placeholder. So the fix is now: make the placeholder a containing block and fill it with `inset: 0`. No `anchor()`, no `.positioned` gating, no `node_visibility` change. Title and description updated to match.

michael:
That looks good. Merging.

PR #287, opened 2026-05-20
Scroll into view fixes during node selection rerenders

johannesmutter:
Enter on a node gap ran node_at_cursor.scrollIntoView({block:'nearest'}) in svedit's __render_node_selection, which jumps the page whenever the adjacent node is taller than the viewport (nearest can't no-op on an element that's never fully visible). Solution: skip that scroll when the cursor's edge is already on screen.

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -1151,6 +1151,22 @@ ${fallback_html}`;
 		);
 	}
 
+	/**
+	 * True when any part of the element's border box overlaps the window
+	 * viewport. Used to keep node-selection re-renders from scrolling a
+	 * cursor that is already on screen. Tests the window viewport, so a
+	 * node clipped only by an inner scroll container still counts visible.
+	 * @param {Element | null | undefined} el
+	 */
+	function __intersects_viewport(el) {
+		if (!el) return false;
+		const r = el.getBoundingClientRect();
+		return (
+			r.bottom > 0 && r.top < window.innerHeight &&
+			r.right > 0 && r.left < window.innerWidth
+		);
+	}
+
 	function __render_node_selection() {
 		const selection = /** @type {NodeSelection} */ (session.selection);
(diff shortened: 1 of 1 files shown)


michael:
Yeah I think that's some regression of https://github.com/michael/svedit/pull/278, was also already investigating. 

I'd definitely expect that if a node is partially in view already there should be no scrollIntoView happening.

Let me check if this fixes it for me....

michael:
It's better, but still overfiring. I'd like to keep the scroll position where i am currently if the node is already partially in view (after a transaction ran).

Behavior with Svedit 0.9.0:


https://github.com/user-attachments/assets/d78508eb-c136-458c-be0d-0c7037a70857


Current behavior (this branch)


https://github.com/user-attachments/assets/7b25440d-7b06-4e22-aa13-a3d21cb2ce47



For the same reason i think we trigger an unexpected scrollTo, when you have a node selection, and then your window looses focus. And then when it regains focus, it scrolls (which is unexpected)

johannesmutter:
I can't really see what's going on in the videos. are you pressing some keyboard keys?

michael:
Yes, I'm cycling through types/layouts. 

But the general issue is this: 
- There's a node selection
- Some transaction runs (e.g. change type/layout)
- Node selection is rerendered (node is already partially in view, but scrollto still triggers, but it shouldn't to keep things calm/stable)

Here's another reproduction of the same issue (quite sure):
- scroll to a story, so that it is half cut-off at the bottom
- select the title
- press esc (which will select the parent node)
- now it scrolls the node into view (but there should be no scrollto because the node already is partly visible)

And a third one:
- select a full story that is partly in view (bottom cut off)
- loose window focus with cmd+tab
- gain window focus again with cmd+tab
- a scrollto happens, unexpectedly

johannesmutter:
All the issues listed are fixed now

michael:
Much better!

Issue #289, opened 2026-05-21
Critical: #287 caused a regression where a path with a negative index is computed when position 0 is selected


michael:
Reproduce on svedit.dev
- Create a first story button (cursor is at pos 1)
- Remove the button (caret is at pos 0 -> error is thrown)

<img width="1224" height="662" alt="Image" src="https://github.com/user-attachments/assets/477ab7cb-bb69-42f9-8223-d1bfafdd1422" />

johannesmutter:
Fixed in https://github.com/michael/svedit/pull/290

Issue #288, opened 2026-05-21
Scroll node into view - sometimes missed


michael:
There are still situations, where a node is selected (that is out of view completely) 

Reproduce:
- Select the "Minimal viable editor" story node
- Delete it
- Scroll to the bottom of the page
- Undo (node selection is not revealed on screen)

https://github.com/user-attachments/assets/0b4e08ad-d5b0-4c0c-af06-bc5c67ead6d1

In a similar situation it's scrolling "very close" to the node, but still not making it visible. 

Similar reproduction:
- Select the "Nested nodes" story node
- Delete it
- Scroll to the bottom of the page
- Undo (node selection is not revealed on screen)

https://github.com/user-attachments/assets/5ec76bc0-ce77-4382-8516-5af4c8c26996

Hard to say what the difference in these two situations is (like why is it off more in the first situation). Maybe a timing issue. However I think this is the last UX issue related to "node scrolls" - we're close.

michael:
With collapsed node cursors there are similar issues. Try this:

- select the last node gap in body
- press backspace 10 times, making short breaks in betweeen

=> sometimes the blinking node cursor is revealed (right at the edge), sometimes not.

johannesmutter:
> Reproduce:
> 
> * Select the "Minimal viable editor" story node
> * Delete it
> * Scroll to the bottom of the page
> * Undo (node selection is not revealed on screen)

fixed in https://github.com/michael/svedit/pull/290

> With collapsed node cursors there are similar issues. Try this:
> 
> * select the last node gap in body
> * press backspace 10 times, making short breaks in betweeen
> 
> => sometimes the blinking node cursor is revealed (right at the edge), sometimes not.

Fixed in https://github.com/michael/svedit/pull/290

johannesmutter:
The remaining multi-node case from (select-all buttons → delete → scroll away → undo) turned out to be a separate, deeper bug from the fixes in this PR, so I'm putting it now on its own branch: fix/node-scroll-after-relayout. It's a timing issue.

johannesmutter:
Fixed via: https://github.com/michael/svedit/pull/291

PR #290, opened 2026-05-21
Fix node-selection scroll-into-view crash and wrong target

johannesmutter:
Fix for the issues #289  and #288 

---

### 289 — crash when a node caret sits at offset 0 (critical)

A regression from #287. It added an unconditional lookup:

```js
const node_before = __get_node_element(node_array_path, cursor_offset - 1);
```

When the caret is at the first gap (`cursor_offset === 0`), `cursor_offset - 1` is `-1`, and `serialize_path` throws `Path index must be a non-negative integer: -1`.

Fixed by skipping the lookup at offset 0:

```js
const node_before = cursor_offset > 0
    ? __get_node_element(node_array_path, cursor_offset - 1)
    : null;
```

### 288 — off-screen node selection scrolled to the wrong node

The general case scrolled `node_after` — the node *after* the selection (`node[cursor_offset]`). For a node-range selection the selected node is `node_before` (`node[cursor_offset - 1]`), so the scroll revealed the following node and left the selection itself off-screen above. The miss equals the selected node's height — which is why a taller story is "off more".

Fixed by scrolling the selected node for range selections; a collapsed caret keeps `node_after`, whose leading edge is the cursor:

```js
const target = is_collapsed ? node_after : node_before;
target?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
```

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -1230,8 +1230,12 @@ ${fallback_html}`;
 
 			// If either node flanking the cursor is already (even
 			// partially) on screen, the cursor is visible — keep the
-			// viewport stable and scroll nothing.
-			const node_before = __get_node_element(node_array_path, cursor_offset - 1);
+			// viewport stable and scroll nothing. node_before is null at
+			// offset 0: cursor_offset - 1 would be -1, and serialize_path
+			// rejects a negative index.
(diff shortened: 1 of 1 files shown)


michael:
Works!

PR #291, opened 2026-05-21
Retry node-selection scroll until the cursor is on screen

johannesmutter:
**Problem.** After an undo/insert, the restored nodes' layout (images, content sizing) settles over several frames. `__render_node_selection` scrolled once, at a fixed-too-early moment, against an un-settled layout — leaving the cursor off-screen. Repro: select all buttons in a story → delete → scroll away → undo; previously only a *second* undo revealed them, once layout had settled.

**Fix.** The render no longer scrolls once and gives up: if the DOM isn't mounted yet it retries on the next frame, and the scroll re-runs each frame until the cursor is genuinely in view (bounded to ~1s).

**Why this layer.** Earlier attempts each baked in a timing assumption — `setTimeout(0)`, a 500ms observer window, leading-edge visibility — and each was wrong in some case. This assumes nothing: it's defined by the *outcome* ("is the cursor in view?") and converges regardless of when or why layout settles. It tolerates the async layout rather than eliminating it — which is correct, since indeterminate content size until load is inherent to the platform. Going deeper would mean deterministic layout from the model (e.g. media nodes carrying dimensions) or an unbounded event-driven `ResizeObserver` instead of the bounded poll — both optional refinements, not corrections.

Resolves the remaining multi-node case from #288.

### DIFF
--- a/src/lib/NodeArrayProperty.svelte
+++ b/src/lib/NodeArrayProperty.svelte
@@ -102,4 +102,23 @@
 		min-width: 24px;
 		cursor: pointer;
 	}
+
+	/* Reserve room at the scroll edges so the browser's caret-follow scroll
+	   (when typing at the end of an overflowing node) can't pin the caret
+	   flush against the array's clip edge. Without this padding, scrollLeft
+	   stops just short of `scrollWidth - clientWidth - EDGE_TOLERANCE_PX`,
+	   leaving `edge_map.last` false; the trailing gap never gains
+	   `.positioned`, and the user sees the caret at the edge with no gap to
+	   click. The default is 2 × edge-gap (≈ enough for the gap plus a
+	   click margin); override per-array via the custom properties. */
+	:global([data-type="node_array"]) {
+		scroll-padding-inline: var(
+			--node-caret-scroll-padding-inline,
+			calc(var(--node-caret-edge-gap, 24px) * 2)
+		);
+		scroll-padding-block: var(
+			--node-caret-scroll-padding-block,
+			calc(var(--node-caret-edge-gap, 24px) * 2)
+		);
+	}
 </style>
--- a/src/lib/NodeGapMarkers.svelte
+++ b/src/lib/NodeGapMarkers.svelte
@@ -94,10 +94,14 @@
 		if (sorted[0] === 0 && edge_state?.first === true) {
 			add_offset(0);
 		}
-		// Mid gaps: between consecutive visible nodes. The gap at offset N
-		// sits between node N-1 and node N — emit when both are visible.
-		for (let i = 0; i < sorted.length - 1; i++) {
-			if (sorted[i + 1] === sorted[i] + 1) add_offset(sorted[i + 1]);
+		// Mid gaps: anchor to the previous node. Emit a marker whenever the
+		// prev flank is visible, even if next falls outside the overscan —
+		// mirrors the NodeGap gate in node_visibility.svelte.js. Wrap/grid
+		// layouts whose row height exceeds the overscan (e.g. tetris with
+		// tall cells) would otherwise lose every between-row marker even
+		// while the user can see and click the underlying NodeGap.
+		for (const i of sorted) {
+			if (i + 1 > 0 && i + 1 < count) add_offset(i + 1);
 		}
 		// Edge gap-after.last: same gate as gap-before but on the trailing edge.
 		if (sorted[sorted.length - 1] === count - 1 && edge_state?.last === true) {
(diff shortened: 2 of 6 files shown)


michael:
Problem description is spot on. setTimeout(fn, 0) was too naive here.

Fix definitely improves situation. I'm still testing and reproducing as I still found cases where the active node selection isn't revealed. will document as soon as I can reproduce reliably.

We definitely need some notion of "wait until layout has settled" then scroll. Though  we might only get heuristics here. So hard to say which approach is the simplest+most robust.

Experimenting with an alternative approach (#292) and applying it to all scroll to situations (not just node selections)

johannesmutter:
@michael Pushed a1c4ef6 combining both approaches.

Node selection: IntersectionObserver watching the two flanking nodes — re-fires the scroll on every viewport-edge crossing, so the relayout cases that survived the first visibility-pass (your "still found cases") now get caught. Stops on selection change, user scroll (with a 500ms grace so the settling itself can't abort it), or a 2s safety timeout.

Property + text selection: adopted your tick() + (rAF || 50ms) + deferred_scroll_job_id pattern from #292 — those don't have the flank-settling structure, so single-shot is the right shape there. A node-scroll-watch start also bumps the job-id to invalidate any pending property/text scroll queued just before it.

michael:
This looks good on a test within Svedit in Chrome. But since this introduces quite a bit of machinery and event sources, I'd like to take my time to test against this branch with the EW app for a while, and merge once confident enough. Close to releasing 0.10.0, and this will likely be part of a follow-up patch release.

Do you really think binding to wheel and touchmove is necessary here? If I understand correctly this is meant to cancel the automatic scroll once the user interacts. Just my gut says that these additional event sources will cause race conditions. I don't think the scroll-into-view flow should run longer than 500ms, and I think that's an acceptable period to block the UI for scroll input. So maybe the cancel logic isn't needed.

Another concern: The intersection observer assumes that the window viewport is scrollable, but Svedit could also be mounted/rendered inside an overflowing container.

Found another remaining scroll-into-view issue: When you type inside a button right at the right edge of the button container in story, and then use the right arrow or down key, the last node gap will be selected, but not revealed in the viewport.

https://github.com/user-attachments/assets/4e43700c-a779-46e0-893b-e953dd2597ec

johannesmutter:
Another issue I found: set this css anywhere and node gaps will not render at all:
```
content-visibility: auto;
contain-intrinsic-size: 100% 100vh;
```

johannesmutter:
tmp: testing new approach. performance:

| 50    | 46        | 672    | 20   | 28     | 60         | 30         | 60         | 60         | 50           | 30           |
| 100   | 94        | 1352   | 18   | 28     | 59         | 30         | 60         | 57         | 44           | 30           |
| 200   | 148       | 2715   | 18   | 42     | 40         | 20         | 53         | 30         | 32           | 20           |
| 500   | 312       | 6823   | 16   | 61     | 16         | 12         | 19         | 10         | 12           | 8            |
| 1000  | 763       | 13679  | 15   | 84     | 8          | 4          | 4          | 2          | 3            | 2            |
| 2000  | 2460      | 27372  | 15   | 197    | 1          | 1          | 1          | 0          | 1            | 0            |

johannesmutter:
This issue is now addressed by adding css scroll padding:

> Found another remaining scroll-into-view issue: When you type inside a button right at the right edge of the button container in story, and then use the right arrow or down key, the last node gap will be selected, but not revealed in the viewport.

johannesmutter:
@michael I addressed your feedback and a couple other issues I found.

michael:
> content-visibility: auto;
> contain-intrinsic-size: 100% 100vh;

Is this addressed also already? And can you explain what is/was the issue here? And when would you use this? From the docs I found I still don't understand the use-cases and problems it solves. Mainly a performance optimization I guess?

johannesmutter:
RE: content-visibility: auto

`content-visibility: auto` is a performance primitive: the browser skips layout/paint/a11y for off-screen subtrees and renders them as a placeholder sized via `contain-intrinsic-size`. It's how you'd virtualise a long page without writing JS virtualisation. My app used it in a presentation view with hundreds of full height slides (I have adjusted it to now only use it in view mode).

It breaks node gaps because of how CSS anchor positioning interacts with containment. While the cell is off-screen, `content-visibility: auto` applies `contain: size layout`, which (a) replaces the cell's content box with the placeholder size: so any `anchor()` reading the node's edges resolves against the wrong rect and (b) scopes the contained subtree's anchor-name discovery so gaps that live *outside* the cell can't see the node's anchor-name across the boundary. Setting it anywhere kills every gap, not just the ones near the visibility boundary.

So this issue is for another day/ PR. The fix is either: 
- move gaps inside the cell (we had this before, which has other downsides), 
- use `anchor-scope` once browesr support stabilises
- or something else I'm not yet aware of.

PR #295, opened 2026-05-26
Expose anchor-name and gap path/offset on .gap-marker

johannesmutter:
## Summary

Three attributes per marker so consumers can target a specific gap from CSS / JS without re-deriving its geometry:
- `anchor-name: --gm-<key>` — CSS anchor positioning for popovers/menus pinned to a gap
- `data-gap-array-path` + `data-gap-offset` — path-based lookup

No behavior change inside svedit; existing styling and caret logic ignore them.

## Test plan

- [x] svedit-lab smoke tests still pass
- [x] Existing gap-marker appearance and caret behavior unchanged in editor

### DIFF
--- a/src/lib/NodeGapMarkers.svelte
+++ b/src/lib/NodeGapMarkers.svelte
@@ -147,7 +147,9 @@
 		class:first={gap.is_first}
 		class:last={gap.is_last}
 		class:pair={gap.has_pair}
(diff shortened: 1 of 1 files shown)


johannesmutter:
@michael can we land this please soon in main (it's a blocker for me)

PR #296, opened 2026-05-26
register_mount: refcount + microtask-deferred duplicate check

johannesmutter:
## Root problems

**1. `register_mount` is intolerant of Svelte's mount/unmount ordering.**

`{#key …}` (and other conditional renders) mount the new subtree *before* the old subtree's `$effect` cleanups run. For one microtask, both subtrees coexist and both register the same paths. The current synchronous check fires `console.error` on this transient overlap — a false positive on a legal Svelte pattern.

**2. The check is unreliable even when it does catch a real duplicate.**

When it sees a second register, it bails *without recording it*. The first `unregister_mount` then deletes the registry entry — leaving the second mount unrecorded. Any future duplicate of that same path will not fire the check, because the registry no longer knows the path is mounted at all.

## How it surfaces (real-world example)

A consumer with a large reactive doc (~144 components reading `svedit.editable`) finds that reactively toggling `editable={…}` walks every component — slow (~705ms `tick()`). They wrap `<Svedit/>` in `{#key is_edit_mode}` to take the mount path instead (~325ms). Every toggle then spams 50+ `[svedit] Path "…" is mounted more than once.` errors. Final DOM state is correct — just noise.

## Fix

- `#mounted_paths`: `Record<string, true>` → `Map<string, number>` (refcounted)
- `register_mount` queues a one-shot microtask check when count goes > 1; logs only if count is still > 1 after Svelte drains pending cleanups
- A transient swap-window overlap unwinds before the check fires → silent
- A persistent overlap (header + footer rendering the same path) still warns
- The refcount also fixes problem #2 incidentally: the second register is recorded, the first unregister only decrements

## Test plan

- [ ] Mount a path twice synchronously, do not unregister → `console.error` fires after the microtask
- [ ] Mount, mount, unmount in one sync block → no error
- [ ] Real-world: cmd+E toggle on a `{#key …}<Svedit/>` consumer goes from 50+ errors per swap to 0

### DIFF
--- a/src/lib/Session.svelte.js
+++ b/src/lib/Session.svelte.js
@@ -66,35 +66,62 @@ export default class Session {
 	available_annotation_types = $derived(this.get_available_annotation_types());
 
 	/**
-	 * Tracks paths currently mounted in the DOM. Within one Svedit document, each
-	 * path may be mounted exactly once. Used by NodeArrayProperty to detect
-	 * duplicate mounts in dev mode.
-	 * @type {Record<string, true>}
+	 * Tracks paths currently mounted in the DOM, by reference count. Within
+	 * one Svedit document a given path may be mounted exactly once at steady
+	 * state — but a transient overlap of 2 (one mounting, one unmounting on
+	 * the same microtask) is legal during a consumer-driven swap, e.g. a
+	 * `{#key …}<Svedit … />{/key}` flip where Svelte builds the new subtree
+	 * before the old one's $effect cleanups run. We refcount the mounts and
+	 * defer the duplicate check to a microtask so genuine overlaps that
+	 * resolve themselves don't fire a false-positive console error.
+	 * @type {Map<string, number>}
 	 */
-	#mounted_paths = $state.raw({});
+	#mounted_paths = new Map();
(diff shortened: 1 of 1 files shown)


michael:
Have a feeling the root issue here is an ownership problem. Probably Svedit should not register mounts on the session (which lives outside the component lifecycle), but on the svedit instance (=component). Then the mounts wouldn't mix up in one registry I guess. Need a bit of time on this, but will look into it.

michael:
Honestly, maybe the whole mount registry is not worth it. It's just a developer safety net, to get an error once you try mount multiple components to the same path. It also consumes quite a bit of resources, just to have that runtime check during development. So maybe just documenting this in the README is enough. You'd anyways notice some weird behavior, like loosing the selection if you display the same path multiple times.

I guess we should just let Svelte do it's reconciliation work, and don't get involved with it.

Like we were creating more problems here than we solved.

michael:
Fixed via https://github.com/michael/svedit/pull/301. Gives the user similar errors, without having to hook into component lifecycle, which created more problems than it solved.

PR #315, opened 2026-06-23
Allow modifier keys only on physical keyboards and ignore on virtual keyboards

michael:


michael:
See: https://discuss.prosemirror.net/t/shift-enter-in-mobile-phones/4903/6

PR #316, opened 2026-06-24
Introduce marks and annotations for node_array

michael:


johannesmutter:
Hey Michael,
Saw your tweet on node-array annotations and had to dig in, because I built two flavors of "grouping" into Interface Lineage a while back and was curious how they line up with this native version. 

Sharing my thoughts in case it's useful for Editable Website and because I'm proposing a change regarding "overlapping" annotations:

I ended up with three places an "anchored" property can live, and I think of them physically:
- **Room**: a real nested node array (a group). Walls; you go *inside*; it nests N levels deep. This is where layout lives: stack becomes columns, direction and spacing change. Container and geometry.
- **Zone**: your new node-array annotations. A painted region on the floor, no walls. The blocks stay in the parent flow; they're *covered*, not contained. Flow membership: reading order and direction never change.
- **Sticker**: a property on a single node. It travels with the node when you lift it out of any room or zone. For things that belong to one block, not a run.

So **annotations** for things that live in the reading flow (direction never changes) and **groups** for things that get placed (direction and spacing usually do).

The distinction underneath all three is what the data is "anchored" to:

- A **sticker** is anchored to the node's **identity**, so it travels. Home for intrinsic facts (shared or specific to one type only)
- A **zone** is anchored to a **position in a container**, so it's contextual: move the block out and the marking stays behind, pointing at the span of nodes, not the node.
- A **room** is a **reified unit**: the only one you can grab, name, nest, or lay out as a single thing.

So in total 4x attachment points now exist:

1. text annotation: a character range (contextual to the text)
2. node property: travels with the node (sticker)
3. node-array annotation: a block range (zone)
4. group: a container that reifies and nests (room)

I think as editor authors, being opinionated about *which to surface when* is the whole game. So the questions I route with are (maybe a guide to add to the readme? let me know and I'll make a PR):

- Should the property survive being moved out, or vanish? → survive = sticker; vanish = zone or room
- Do I need to grab the collection as a unit (move / name / nest / lay out)? → room
- Or will I grab blocks individually and move them around the doc? → zone or sticker
- Is it geometry, or is it paint? → geometry = room; paint = zone

---

**The one thing I'd push on: drop exclusivity for node arrays.** For text it's defensible. For blocks, the things you want to mark are *layers* that must coexist: this run is a section **and** has a comment **and** is flagged for review **and** is machine-translated. One-at-a-time is far more limiting on blocks than on characters, where you rarely want two coexisting section-like concepts.

I think exclusivity only feels necessary because of the rendering choice. One `<section>` div around the run forces the "which wrapper is outer?" tree problem, and very importantly it quietly breaks flatness (and most layouts!): covered blocks stop being siblings, which bites the instant the parent is a grid or flex wrap layout (they collapse into one item). 

But if you render **per node** instead: each covered block just knows which annotations cover it. You already pass `is_start/is_middle/is_end`, so make it a *list*. Then:
- overlap is free — a block carries several tags at once, no box to nest;
- flatness holds — no wrapper, blocks stay direct children, which is actually *more* in line with Svedit's "same DOM in both read/edit modes" principle than the spanning div is;
- ranges stay exactly as-is for authoring and storage; you just resolve range → per-node membership at render. Store as ranges, render as tags.

And I don't think you lose anything, because the case that seems to want a wrapping box, a bordered card with padding around the run, isn't really an annotation: it has walls, so it's a nested node array (a group). Annotations are paint on the blocks that happen to be covered; groups are the box. If we keep that line sharp then node-array annotations get dead simple: always per-node, always overlappable, no e.g. "wrap mode", no flag.

---

Aside note: One thing that clarified it for me: overlap doesn't remove the need for node properties, it *purifies* it. Before overlap, node properties were doing two jobs, holding genuinely intrinsic data *and* being a workaround for "I need more than one marking and annotations are exclusive." Overlap hands that second job to zones, and the sticker is left with only its real job: identity-anchored data that travels. So the two axes are clean and orthogonal: overlap is about *multiplicity* (how many markings per block), sticker-vs-zone is about *anchoring* (does it travel). They don't compete.

Anyway, really nice to see this land natively. Let me know if you want to discuss the overlapping topic or if should to a PR that branches of this one that has this implemented.

johannesmutter:
one addition:
the zones or node array annotations have solved for me the "overlapping node gaps" issue in almost all cases, because groups (nested node arrays) there's always a spacing and annotations have no additional wrapper so no additional node gap. So that's a nice side effect ✌️

michael:
Regarding overlapping annotations: That was not in the scope here, and I'm still not sure if they fit my vision for Svedit. But maybe. There are situations where we already need it like the selection fragment (for anchoring) can currently only be rendered if there's no annotation in the way. And there are user requests to stick additional annotations to text, like comments. Those requests are valid as yours to allow overlapping annotations for node arrays. Again, it's out of scope for this next release, but definitely a limitation of Svedit atm.

Text and node (array) annotations are designed to be conceptually the same in Svedit. That's the model I'm following at the moment, and there's shared code, and lots of consideration regarding common annotation toggling behavior. Tried to find a sweet spot here, without introducing a configuration layer for toggling behavior. For example if you create a Strong annotation that touches another Strong annotation, you may just extend the existing one. That would be cleaner. But that doesn't work everywhere. E.g. it's valid to put two links that are touching each other, so they should not "unite". Also the sections that can be attached to page.body. You should be able to create a new section right after one section ends. For that reason when you type right at the boundary of annotation the new text will not be included or when you hit Enter at a node gap right at a section boundary. The mechanics that are in place now work like this: You select some nodes in page.body. The selection may touch one or more annotations (e.g. sections). The toggleannotation command will be marked as active (if all touched annotations have the same type). Executing it will delete all the annotations. Toggling again will create a new annotation exactly at the selection you made. I found that this works quite well for the section grouping use case, as well as for the text annotations. With 2 clicks you'll get a clean annotation at your selection range. There's no expand/truncate logic (that I first thought, I needed). Just delete+recreate. Expand/truncate logic may still make sense for all annotations that carry data, beause those you don't want to destroy and recreate. But that's for another release.

I was considering just introducing a section node for grouping. But as you say, that creates "walls" and you can no longer select across section boundaries, it also creates additional node gaps. Annotations for grouping feel softer and much better here. 

Regarding the wrapping HTML element that represents a node annotation: I was originally expecting that I'm forced to not render them, because they might likely break assumptions of the node gap rendering. But to my surprise it worked just fine, and selections are also correctly mapped despite the extra markup. So I can have the API be exactly the same as for text annotations. Section.svelte renders a `<section>` and you can for instance set a background color there. Just you'd render a `<strong>` element as text annotation. Of course this works only when annotations are exclusive, which is perfectly fine for me at the moment as a restriction.

However, I'm open to revisit this in the future, if a generalized solution exists that works the same for text and node annotations. Your suggestion to just keep node arrays flat and pass down `annotations` to its child doesn't work for text annotations. You need markup there and a fragmenter solution. E.g. ProseMirror does this:

```html
<p>
  <strong>Strong</strong>
  <em>
    <strong>AndEmphasis</strong>
  </em>
</p>
```

Not good in our case, because then you'd have two strong elements, and anchoring becomes ambiguous for instance. So a bunch of hard problems that are maybe not worth tackling. Possibly I want to position Svedit deliberately as a content-only tool, and not as a foundation for productivity tools where annotations matter a lot. But I'm not not sure yet where the journey goes. For now it's an okay limitation.

johannesmutter:
Just to clarify: I'm not arguing for overlapping annotations on text, only on node arrays. And no pressure on any of this.

On the comments case: I keep the comment out of the inline layer that already holds highlights, bold, italic, and links. The comment-count trigger renders in the _margin_ beside the text. Two things fall out of that: the visual layers stop competing, and links and comments can never contend for the same hitbox. So the "overlap gets messy" worry is a rendering decision, not a reason to forbid overlap in the model. The annotation still overlaps in the data; only its indicator moves to the margin.

The core limitation of wrapping a node run in a `<section>`: it only holds when the array stacks in one direction. Make the parent a grid or flex-wrap and the wrapper becomes a single grid item, collapsing everything it covers into one cell. It worked in your tests because a stacked column is the friendly case. Text sidesteps this entirely: a text annotation is inline, and inline boxes fragment across line breaks for free. A block-level wrapper can't fragment across grid tracks. In the Svedit demo that's exactly why the image-grid node array can't take an annotation.

That asymmetry is why I'd not force text and node-array annotations to render the same way. Keep the range model identical for authoring and storage. Resolve range to per-node membership at render, so each block carries the list of annotations covering it. The grid stays intact and overlap comes for free, with nothing about text touched.

> Your suggestion to just keep node arrays flat and pass down annotations to its child doesn't work for text annotations

Agreed. I'm not proposing that for text. Text needs the markup and a fragmenter. Per-node is only for node arrays.

Recording of adding comments (on top of e.g. other annotations, but not visually or in markup overlapping, only in data)

https://github.com/user-attachments/assets/bcf52c47-07c9-4261-8053-0fe56957f175

michael:
Idea: What if you can opt-out of rendering annotations in-place by simply not registering a component for those? Which would mean that exclusivity of annotations only needs to be enforced for all annotations that render in-place, not for those that just exist on the data layer, and you render elsewhere (like your comments in the margin of the owning node). 

This would also allow you to opt-out of rendering a wrapping element in node arrays, so it's up to you. Want a wrapper for sections? Register `section: Section`. For gallery.items that renders with a grid, leave it out.

Downsides:
- it's hard to maintain those "unmanaged" annotations because you can not visually anchor them. Like you're a bit blind. Even in your comments-on-the side case. You may delete the word in the text where you anchored your comment originally, and be surprised why the comment count went down. Like why have comments anchored to a text, if you can't render it there? Wouldn't it be better if you just add a comments property on the node that owns that text property?
- error prone: we'd need to trust that users create and delete those "invisible" annotations reliably. Say you create overlapping comments annotations one day, but register a component Comment another day, you'd break rendering.
- you can not utilize ToggleAnnotationCommand, and tr.toggle_annotation, which implement "exclusive behavior" so you have to use tr.create/set etc. yourself. 
- or if you want support from the library you'd have to introduce annotation kind: 'execlusive|non-exclusive' or some ignore_exclusivity flag for tr.toggle_annotation but that's a receipe for data corruption i have a feeling.

TL;DR: Maybe I can "prepare" for a future where an exclusive (renders in place) and a non-exlusive layer (u are responsible for rendering outside) can co-exist. By making annotation rendering optional. But for now I'd still document/and enforce exclusivity in the docs. But you and your app can break out of that if you take responsibility for maintaining those invisible annotations. Maybe there can be better API support later, and safeguards, just it's not clear to me yet how that should look like and i need use cases of my own to properly think this through.

michael:
Pushed a commit that makes annotation rendering optional, which gives you some flexibility here.

https://github.com/michael/svedit/pull/316/changes/eaed628cf413b205f38517675cf9f4a6879cfe1c

johannesmutter:
The exclusivity / no-overlapping annotations limitation comes from the issue that you cannot easily overlap dom nodes, or was there another reason? Like being opinionated about the design: not mixing bold, italic, color, background color?

e.g. this is invalid:
```html
<em>italic and <b>bold and italic</em> and bold</b>
``` 
it would require `html <em>italic and <b>bold and italic</b></em><b> and bold</b>`
<em>italic and <b>bold and italic</b></em><b> and bold</b>

But css highlights don't have that limitation:
```javascript
const x = new Range();
x.setStart(parentNode, 55);
x.setEnd(parentNode, 62);
const highlight = new Highlight(x)
```
`<em>italic and <b>bold and italic</b></em><b> and bold comment</b>`
<em>italic and <b>bold and italic</b></em><b> and bold <mark>comment</mark></b>

> it's hard to maintain those "unmanaged" annotations because you can not visually anchor them. Like you're a bit blind. Even in your comments-on-the side case. You may delete the word in the text where you anchored your comment originally, and be surprised why the comment count went down. 

So you CAN have a visual anchor. You're however limited to what css highlight api supports.

---

> Like why have comments anchored to a text, if you can't render it there? Wouldn't it be better if you just add a comments property on the node that owns that text property?

In my case I preferred no visual overlay to keep it cleaner, especially if you have dozens of comments over the same text. And for a longer text in the same text node it's kind of crucial to know which words it referred to. Though I think in the generals scenario yes you would want to visualize those highlights, maybe on hover only or when the corresponding comment is active.

johannesmutter:
The guard you added in principle makes already sense (incase one forgot to register that component), but it feels a bit hacky to check on `svedit.session.config.node_components[fragment.node.type]` wether to render the wrapper around nodes or not. Because now if want to wrap individual nodes instead with an annotation driven by node array annotations, instead of wrapping fragments, I can't really register a component in config, I need to do another hack.

michael:
> The exclusivity / no-overlapping annotations limitation comes from the issue that you cannot easily overlap dom nodes, or was there another reason? Like being opinionated about the design: not mixing bold, italic, color, background color?

Both: I like the simplicity of exclusive annotations. It's more semantic/precise than allows users to mix bold/italic/underline and slam a link on top of all of it. So i'd like to keep the exclusivity on the "content" layer.

For the annotation layer (comments etc.) it might be okay to break that rule, though I don't want Svedit to be responsible for all possible behaviors of overlapping annotations. So I'm okay to enable it but then leave all the management to the app. Like you need to decide how to render, how to create, update and delete those pieces of data. It's just important that they are ignored in the render path of Svedit's NodeArray and TextProperty.

Here's an attempt to expose a simple API: https://github.com/michael/svedit/pull/319 

Child nodes of a node array get an `annotation` prop which is only set if there is exactly one annotation (this will cover all my use cases where excuslivity is on). And then there's a new `annotations` prop that contains all the annotations, like might be 100 comments that are overlapping.


> The guard you added in principle makes already sense (incase one forgot to register that component), but it feels a bit hacky to check on svedit.session.config.node_components[fragment.node.type] wether to render the wrapper around nodes or not. Because now if want to wrap individual nodes instead with an annotation driven by node array annotations, instead of wrapping fragments, I can't really register a component in config, I need to do another hack.

I think this is fine. The node_components registry is there so NodeArray and TextProperty can delegate to their child components. You are not in control of the full rendering path here. Only indirectly through the registry. If you take control of rendering annotaitons elswhere e.g. somewhere in Overlays.svelte, then you can just import components directly, or create your own registry.

Have a look at PR if that would be helpful for you? I don't want to add more stuff at this point, just unblocking the use cases you described.

michael:
If you want TextProperty to take care of CSS highlights, you can just roll your own custom TextProperty.svelte to do that.

johannesmutter:
Setting overlap aside for a moment, because the limitation I keep coming back to is independent of it: **with a fragment wrapper, node-array annotations only work in layouts that stack in one direction.** The demo already contains the counter-example: `ImageGrid` renders its node array with `display: grid`. Cover two images with any annotation — even an exclusive one — and the `<section>` wrapper becomes a single grid item: the covered images collapse into one cell and stack inside it. Text never hits this because inline boxes fragment across line breaks for free; a block-level wrapper can't fragment across grid tracks. So I think text and node annotations can share the model (which I like a lot), but they shouldn't be symmetric at the render layer … and the fragmenter problems you describe (ProseMirror-style splitting, ambiguous anchoring) are text-layer problems that node arrays never need to inherit.

The opt-out commit helps, but it couples "no wrapper" to "no component". As it stands I can have either:

- a **visible** annotation → non-wrapping layouts only, or
- a **grid-safe** annotation → invisible, with exactly the "unmanaged, blind, error-prone" downsides you listed.

The middle option is what I'm actually proposing: **keep the component, mount it per covered node instead of around the run.**

| | Fragment mount (current) | Per-node mount (proposed) |
|---|---|---|
| Component instances | one, wrapping the run | one per covered node, receiving the existing `is_start` / `is_middle` / `is_end` context |
| Stacked layout | ✅ continuous box for free | ✅ continuous box via `is_start` / `is_end` styling |
| Grid / flex-wrap | ❌ run collapses into one item | ✅ every node keeps its own cell |
| Overlap (if ever wanted) | ❌ invalid nesting → needs a fragmenter | ✅ plain nesting per node, always valid HTML |

**Fragment mount** — annotation over A+B in a 3-column grid:

```html
<div class="grid">
  <annotation>              <!-- becomes ONE grid item -->
    <NodeA /> <NodeB />
  </annotation>
  <NodeC /> ...
</div>
```

<table><tr><td><mark>A<br>B</mark></td><td>C</td><td>D</td></tr><tr><td>E</td><td>F</td><td></td></tr></table>

**Per-node mount** — same annotation, same data:

```html
<div class="grid">
  <annotation><NodeA /></annotation>   <!-- is_start -->
  <annotation><NodeB /></annotation>   <!-- is_end -->
  <NodeC /> ...
</div>
```

<table><tr><td><mark>A</mark></td><td><mark>B</mark></td><td>C</td></tr><tr><td>D</td><td>E</td><td>F</td></tr></table>

Nothing else changes: range model, storage, delete+recreate toggling, exclusivity … all identical. `ToggleAnnotationCommand` works untouched. And because the annotation stays visible and visually anchored, none of the unmanaged-annotation downsides apply.

On the config side, I'd make the choice explicit rather than inferring behavior from whether a component happens to be registered (registration-as-behavior-switch is what felt hacky to me above). Sketch (shape is bikesheddable ***):

```js
annotation_components: {
  section: { component: Section, mount: 'fragment' }, // today's behavior
  comment: { component: Comment, mount: 'node' },     // grid-safe
  translation: {}                                     // data-only, no rendering
}
```

---

But to be clear: Per-node mounting is worth it even if exclusivity stays forever, because it's what makes node-array annotations usable on *any* node array, not only the ones that happen to stack.

---

*** bikeshedding

johannesmutter:
looking at your new PR now 👍

johannesmutter:
*** … and when the "box" also needs to change spacing or direction inside, that's usually the signal it's geometry rather than paint: i.e. a nested node array, not an annotation. But for a purely visual box like your sections, per-node styling covers it.

johannesmutter:
ok this looks promising 👌 


https://github.com/user-attachments/assets/37e2fa4c-8bf8-45e1-b62e-5fc94940550d


the beauty here is that it's again entirely driven by css, e.g. using the --row variable

johannesmutter:
I'll open a new PR with a few suggested changes

johannesmutter:
New PR added here:
https://github.com/michael/svedit/pull/320

PR #319, opened 2026-07-04
[Experimental] API that allows app-managed annotations that can overlap

michael:


michael:
Implemented via https://github.com/michael/svedit/pull/321

PR #320, opened 2026-07-04
Overlapping annotations refinements

johannesmutter:
Builds on #319. Three small changes — two behavior fixes and one DX improvement. Plus a demo `marker` annotation that shows the feature off.

### 1. `annotation` (singular) is now stable

Previously it became `null` as soon as 2+ annotations covered a node. So adding one data-only annotation (e.g. a comment) over a section silently broke components using `let { annotation: section } = $props()`, even though the section wrapper still rendered.

Now: `annotation` is the **in-place (component-backed) annotation** wrapping the node — exclusivity guarantees there's at most one. Data-only annotations never wrap, so they only appear in `annotations` (plural) and never affect the singular.

### 2. Data-only annotations no longer block toggling

Previously `tr.toggle_annotation('section')` bailed whenever the selection touched an annotation of another type, including data-only ones. In a document with many data-only annotations, that's every selection.

Now: annotations only compete for toggling when they render in-place or share the toggled type. A nice consequence: the standard `ToggleAnnotationCommand` works out of the box for data-only types — no custom transforms needed to create overlapping annotations.

### 3. Styling annotations is now pure CSS

`<Node>` now adds annotation classes to covered node wrappers automatically — `anno-<type>`, plus `anno-<type>-start`/`anno-<type>-end` on the run's first and last node — the same way it already adds `first`/`last`.

Before, supporting an annotation on a node type meant wiring the `annotations` prop into every node component you maintain:

```svelte
import { annotation_classes } from '../annotation_classes.js';
<Node {path} class={annotation_classes(annotations)}>
```

Now that's gone. Supporting a new annotation type is a CSS rule, nothing else:

```css
.anno-marker {
	background: color-mix(in oklch, var(--app-primary-fill) 13%, transparent);
}
```

The `annotations` prop stays for components that render something richer than styles (badges, margin indicators, etc.).

### Demo: `marker`

A data-only highlighter for blocks (⌘⇧M or toolbar). No component registered, so it overlaps sections and works inside the image grid. Its entire implementation is one schema entry, one `ToggleAnnotationCommand`, and one stylesheet — the rounded caps and rail outline follow the container's flow direction via the existing `--row` variable. The nine demo node components are untouched.

### Tests

Updated the singular-prop expectations, added two toggle tests (component-backed type across data-only overlap, and the reverse), and a test for the automatic wrapper classes.


johannesmutter:
Screenshot 2026-07-04 003175 Svedit - A tiny library for building editable websites in Svelte@2x

michael:
> Previously it became null as soon as 2+ annotations covered a node. So adding one data-only annotation (e.g. a comment) over a section silently broke components using let { annotation: section } = $props(), even though the section wrapper still rendered.

> Now: annotation is the in-place (component-backed) annotation wrapping the node — exclusivity guarantees there's at most one. Data-only annotations never wrap, so they only appear in annotations (plural) and never affect the singular.

annotation being null when there’s 2+ annotations was intentional. Consider it a shortcut, that is only available when you know there can be just one annotation at a time. As soon as you have more you need to use the plural annotations, which has the complete info to reason about. 

You can not assume your one exclusive annotation is rendered. Perfectly valid to have a section annotation without the wrapper. `annotations` always need to have all annotations, regardless if rendered or overlapping etc. You’re on your own inspecting that.

The other things i need to look at later when at the computer.

michael:
> Previously tr.toggle_annotation('section') bailed whenever the selection touched an annotation of another type, including data-only ones. In a document with many data-only annotations, that's every selection.

> Now: annotations only compete for toggling when they render in-place or share the toggled type. A nice consequence: the standard ToggleAnnotationCommand works out of the box for data-only types — no custom transforms needed to create overlapping annotations.

Again data-only does not mean non-exclusive. And exclusive doesn't mean it’s always rendered in-place. I still want to be able to have exclusive toggle behavior for sections even when I dont render a component for it. 

I'm no sure the relevant_annotations() check is generic enough, as expectation may be app-specific. Currently if you'd try to toggle a comment where there already is a comment it would untoggle the existing comment first. Is that what you want? Is that what most/every app wants? If yes we could try to generalize the ToggleAnnotation command by these rules:

- different types of in-place rendered annotations can not stack/overlap
- same type of any annotation can not stack/overlap

I think that's what you implemented.

Potential edge case (maybe not covered right now?): I want to have a section annotation (but no component) and I want it to be completely exclusive. Can I achieve that by allowing only section annotations (annotation_types: ['section'] for page.body)?

> Demo: marker
A data-only highlighter for blocks (⌘⇧M or toolbar). No component registered, so it overlaps sections and works inside the image grid. Its entire implementation is one schema entry, one ToggleAnnotationCommand, and one stylesheet — the rounded caps and rail outline follow the container's flow direction via the existing --row variable. The nine demo node components are untouched.

I like the demo. The extra classes can be helpful, though you might end up with a lot of them. Tailwind users may benefit less from that as CSS users, but I guess it's a useful addition, much like the .positioned etc. classes.

michael:
Unrelated request. Do you have ideas for better and stylistically matching icons for these five fellas:

image

michael:
> Currently if you'd try to toggle a comment where there already is a comment it would untoggle the existing comment first. Is that what you want? Is that what most/every app wants?

Hmm.. I guess it could be a valid default, and for everything else you need to make custom commands and transforms potentially.

michael:
Did more thinking on this, in seek of a clearer mental model. Existence of components for exclusivity is indeed to implicit and would just be temporary API. I'm currently leaning towards introducing two different layers.

- Marks that are rendered and are exclusive (marks are actual content)
- Annotations: those are like overlays on top of the document. They must not render markup (so we avoid the fragmentation stacking problem, and they can overlap with marks but also with annotations)


WIP Implementation PLAN:

# Split Marks and Annotations

## Summary

Replace the current single `annotations` array with two explicit range arrays:

```js
{
	content: 'Hello world',
	marks: [],
	annotations: []
}
```

Use:

- `marks` for content-level ranges such as `strong`, `emphasis`, `link`, `section`
- `annotations` for metadata/overlay ranges such as comments or markers

Default behavior:

- marks are globally exclusive within a property;
- annotations are exclusive only per annotation type when using `ToggleAnnotationCommand`;
- apps may still stack same-type annotations by using lower-level APIs.

## Schema API

Use separate schema keys on text and node-array properties:

```js
content: {
	type: 'text',
	mark_types: ['strong', 'emphasis', 'link'],
	annotation_types: ['comment'],
	allow_newlines: true
}
```

```js
body: {
	type: 'node_array',
	node_types: ['paragraph', 'story'],
	mark_types: ['section'],
	annotation_types: ['comment']
}
```

Use separate node kinds:

```js
strong: {
	kind: 'mark',
	properties: {}
}

comment: {
	kind: 'annotation',
	properties: {}
}
```

`kind: 'mark'` and `kind: 'annotation'` use the same range shape, but different validation/rendering/editing semantics.

## Value Shape

Text properties:

```js
{
	content: 'Hello world',
	marks: [{ start_offset: 0, end_offset: 5, node_id: 'strong_1' }],
	annotations: [{ start_offset: 2, end_offset: 8, node_id: 'comment_1' }]
}
```

Node-array properties:

```js
{
	nodes: ['paragraph_1', 'paragraph_2'],
	marks: [{ start_offset: 0, end_offset: 2, node_id: 'section_1' }],
	annotations: [{ start_offset: 1, end_offset: 2, node_id: 'comment_1' }]
}
```

## Key Implementation Changes

- Rename existing annotation-range helpers internally to neutral range helpers where useful.
- Validation:
  - `marks` must be mutually exclusive.
  - `annotations` must have valid bounds/references/allowed types, but may overlap at document-validation level.
  - `mark_types` may only reference `kind: 'mark'`.
  - `annotation_types` may only reference `kind: 'annotation'`.
- Rendering:
  - `TextProperty` fragments/renders only `marks`.
  - `NodeArrayProperty` wraps only `marks`.
  - Node-array child components receive:
    - `mark`: the single covering mark, or `null`
    - `annotations`: all covering annotations
- Commands:
  - Implement `ToggleMarkCommand`.
    - Marks are completely exclusive.
    - Touched marks of the same type are removed.
    - A single touched property-less mark may switch to another property-less mark.
  - Implement `ToggleAnnotationCommand`.
    - Only competes with selected annotations of the same type.
    - Touched same-type annotations are removed.
    - If no same-type annotation is touched, creates a new annotation.
    - Does not block on marks or other annotation types.
    - Does not enforce document-level same-type exclusivity.
  - Rename transaction APIs:
    - `tr.toggle_mark(type, properties?)`
    - `tr.toggle_annotation(type, properties?)`
- Selection/session helpers:
  - expose `selected_marks`, `active_mark`
  - expose `selected_annotations`, `active_annotation`
  - ensure mark and annotation helpers are independent
- Demo:
  - move `strong`, `emphasis`, `highlight`, `link`, `section` to `kind: 'mark'`
  - move `marker`/comment-style examples to `kind: 'annotation'`
  - update demo document values to `{ content/nodes, marks, annotations }`
  - wire toolbar mark tools to `ToggleMarkCommand`
  - wire marker/comment-style tool to `ToggleAnnotationCommand`

## Open TODOs Before/While Implementing

- Decide final helper names for shared range logic:
  - likely neutral names such as `range`, `range_record`, `selected_ranges`
  - avoid public names that imply rendering.
- Decide whether marks must have registered components.
  - Pro enforcement: prevents accidentally invisible content semantics.
  - Contra enforcement: marks can still be semantic/content-level even when an app intentionally chooses not to render them.
  - Current recommendation: allow optional rendering for marks, but document the risk clearly.

## Compatibility / Migration

No legacy compatibility layer for runtime editing.

Update all Svedit demo/test docs in-place to the new shape. Existing downstream projects can migrate later.

## Tests

Add/update tests for:

- text and node-array values require both `marks` and `annotations`;
- marks reject overlaps;
- annotations may overlap marks and other annotation types;
- same-type annotations may overlap when created manually;
- `ToggleAnnotationCommand` removes touched same-type annotations instead of stacking;
- `ToggleAnnotationCommand` ignores marks and other annotation types;
- `ToggleMarkCommand` ignores annotations but conflicts with all marks;
- mark rendering still works for text and node arrays;
- node-array children receive `mark` and `annotations`;
- mark and annotation ranges transform correctly on insert/delete;
- paste/cut preserves both marks and annotations;
- old mark usages through `annotation_types` are removed from Svedit tests/demo.

## Assumptions

- `marks` and `annotations` share the same range shape.
- Built-in mark commands are strict; built-in annotation commands are same-type exclusive only by convention.
- Apps that want stacked same-type annotations can use lower-level transaction APIs.

michael:
Would later start an attempt to implement this on top of this branch, keeping the marker demo, but with new API underneath.

Let me know if you see any problems with the proposed API.

Wanted at first avoid splitting up annotations into two groups, but having another subtype like {kind: 'annotation', layer: 'marks' } feels unintuitive, and you'll mix up the different behaviors in your head all the time. I guess that's why it was hard to talk about the behaviors. with the mark vs annotation separation its easier to grasp i found. Terminology would be:

- ranges for text and node_arrays (generic concept that captures marks and annotations)
- and then have separate toggle_mark/annotation transforms and commands

michael:
Fable is on it, with a a refined plan:

# Split marks and annotations

## Context

Svedit currently has a single `annotations` array on text and node_array property values. Recent commits (872b84d, 4c39f71, 4553059) introduced overlapping app-managed annotations, but the semantics are **inferred from component registration**: annotation types with a registered component render in-place and must not overlap; component-less types may overlap and are data-only. That inference is fragile — rendering config silently changes document semantics.

This refactor makes the distinction explicit in the data model:

- **`marks`** — content-level, mutually exclusive, rendered in-place (`strong`, `emphasis`, `highlight`, `link` on text; `section` on node arrays)
- **`annotations`** — metadata/overlay ranges, may overlap, data-only (`marker`, comments)

GPT-5.5's plan was reviewed against the codebase and is sound; this file is the verified/refined version. Decisions confirmed with the user:

1. **`kind: 'annotation'` types must NOT have registered components** — validation error if they do. Marks render in-place; annotations are data-only (CSS classes + props).
2. **CSS classes split**: Node.svelte generates `mark-<type>(-start/-end)` for marks and keeps `anno-<type>(-start/-end)` for annotations.

## Target value shape

```js
// text property value
{ content: 'Hello world', marks: [...], annotations: [...] }
// node_array property value
{ nodes: ['p1', 'p2'], marks: [...], annotations: [...] }
```

Both arrays share the range record shape `{ start_offset, end_offset, node_id }` (rename the `Annotation` type in types.d.ts to `NodeRange`; `Annotation` and a new `Mark` can alias it).

## Schema API

- Property definitions: replace `annotation_types` with two keys — `mark_types` and `annotation_types` (both optional) on `text` and `node_array` properties.
- Node kinds: add `kind: 'mark'`; keep `kind: 'annotation'`. `NodeKind` in [types.d.ts](src/lib/types.d.ts) becomes `'document' | 'block' | 'text' | 'mark' | 'annotation'`.
- Validation: `mark_types` entries must reference `kind: 'mark'` nodes; `annotation_types` entries must reference `kind: 'annotation'` nodes.

## Semantics

- Marks are mutually exclusive within a property (doc-level validation, replacing `validate_in_place_annotations_do_not_overlap`, which is deleted — exclusivity is now decided by kind, not component registration).
- Annotations validate bounds/references/allowed types only; overlap is always allowed at doc level (including same-type — apps can stack via low-level APIs).
- `ToggleMarkCommand` / `tr.toggle_mark(type, props?)`: strict — competes with all marks; touched same-type marks are removed; a single touched property-less mark may switch type (reuse `can_switch_annotation_type` in utils.js, renamed `can_switch_mark_type`).
- `ToggleAnnotationCommand` / `tr.toggle_annotation(type, props?)`: competes only with same-type annotations; touched same-type are removed, otherwise creates a new one; never blocks on marks or other annotation types.

## Implementation changes by file

**[src/lib/doc_utils.js](src/lib/doc_utils.js)**
- Default property values: `{ content: '', marks: [], annotations: [] }` / `{ nodes: [], marks: [], annotations: [] }`.
- `validate_annotations_array` → neutral `validate_range_array(...)`, called for both arrays with the respective allowed-type list and kind check.
- Add mark exclusivity validation using `are_annotation_ranges_exclusive` (renamed `are_ranges_exclusive`).
- Delete `validate_in_place_annotations_do_not_overlap`; add config validation that no `kind: 'annotation'` type has a registered component (decision 1).
- `get_node_array_annotations(value)` → `get_node_array_marks` + `get_node_array_annotations`.

**[src/lib/utils.js](src/lib/utils.js)**
- Rename range helpers to neutral names, applied to both arrays on every text edit:
  - `delete_range_from_annotations` → `delete_range_from_ranges` or `adjust_ranges_for_deletion`
  - `insert_range_into_annotations` → `adjust_ranges_for_insertion`
  - `are_annotation_ranges_exclusive` → `are_ranges_exclusive`
  - `calculate_fragment_ranges` — operates on marks only (rendering)
  - `can_switch_annotation_type` → `can_switch_mark_type`
  - `get_selected_annotations` → split into `get_selected_marks` / `get_selected_annotations` (filter by kind)

**[src/lib/Transaction.svelte.js](src/lib/Transaction.svelte.js)**
- `toggle_annotation` (lines ~438–517) becomes `toggle_mark` with strict semantics (drop the component-registration filtering — all marks compete).
- New `toggle_annotation` with same-type-only semantics.
- All text-mutation paths (insert/delete/split/join) transform both `marks` and `annotations` arrays.

**[src/lib/Session.svelte.js](src/lib/Session.svelte.js)**
- Split derived state: `selected_marks`/`active_mark` and `selected_annotations`/`active_annotation`; `available_annotation_types` → `available_mark_types` + `available_annotation_types` (from the property definition's two keys).

**[src/lib/Command.svelte.js](src/lib/Command.svelte.js)**
- `ToggleAnnotationCommand` → `ToggleMarkCommand` (active/enabled logic against `selected_marks`, `available_mark_types`).
- New `ToggleAnnotationCommand`: enabled when type allowed + expanded selection; active when selection touches a same-type annotation.

**[src/lib/TextProperty.svelte](src/lib/TextProperty.svelte)**
- Fragment/render only `value.marks`; drop `get_in_place_annotations` component filtering (marks always fragment; component-less marks render as a plain wrapper with `mark-<type>` classes).
- Selection-highlight suppression check (line ~45) uses `selected_marks`, not annotations.

**[src/lib/NodeArrayProperty.svelte](src/lib/NodeArrayProperty.svelte)** and **[src/lib/Node.svelte](src/lib/Node.svelte)**
- Wrap fragments by `marks` only. Context provides per-index `mark` (single covering mark or `null`) and `annotations` (all covering annotations).
- Child node props: `mark` + `annotations` (replacing `annotation` + `annotations`).
- Node.svelte class generation: `mark-<type>(-start/-end)` for covering marks, `anno-<type>(-start/-end)` for covering annotations (decision 2).

**[src/lib/paste_utils.js](src/lib/paste_utils.js)** and **[src/lib/transforms.svelte.js](src/lib/transforms.svelte.js)**
- Preserve/merge both arrays on paste, cut, join, split. Empty defaults include both keys.

**[src/lib/types.d.ts](src/lib/types.d.ts)**
- `NodeRange` base type; `AnnotatedText`/`AnnotatedNodeArray` gain `marks`; `NodeProps` gets `mark`/`annotations`; property-definition types get `mark_types`.

**Demo ([src/routes/](src/routes))**
- [create_demo_session.js](src/routes/create_demo_session.js): `strong`, `emphasis`, `highlight`, `link`, `section` → `kind: 'mark'`; `marker` → stays `kind: 'annotation'`. Property defs use `mark_types` (ALL_ANNOTATIONS → ALL_MARKS etc.) and `annotation_types: ['marker']` where applicable. Demo doc values get `marks`/`annotations` arrays. Commands: `toggle_strong/emphasis/highlight/section` → `ToggleMarkCommand`; `toggle_marker` → `ToggleAnnotationCommand`.
- [commands.svelte.js](src/routes/commands.svelte.js): link command uses `selected_marks` / `tr.toggle_mark('link', …)`.
- [Toolbar.svelte](src/routes/components/Toolbar.svelte): mark buttons check `available_mark_types`; marker button checks `available_annotation_types`.
- [styles/annotations.css](src/routes/styles/annotations.css): update selectors for `mark-*` / `anno-*` split.
- README.md: update the annotations section (value shape, schema keys, child-component props) to the new model.

**No compatibility layer.** Update all demo/test docs in place.

## Tests ([src/test/annotations.svelte.test.js](src/test/annotations.svelte.test.js) and friends)

Per project convention, implement test changes but do NOT run tests — the user runs them manually. Update/add:

- Values require both `marks` and `annotations` keys; mark overlap rejected; annotation overlap (incl. same-type, and over marks) accepted.
- `ToggleMarkCommand`: exclusive across all marks, ignores annotations, property-less type switching still works.
- `ToggleAnnotationCommand`: removes touched same-type, ignores marks/other types, no doc-level same-type exclusivity.
- Rendering: text and node-array marks render; children receive `mark` + `annotations`; component on a `kind: 'annotation'` type throws at config validation.
- Range transforms on insert/delete and paste/cut preserve both arrays.
- Update `AnnotationAwareNode.svelte` test component and `create_test_session.js` schemas/docs to new kinds/keys.

## Review notes on the original plan (soundness verdict)

The plan is sound and matches the codebase well. Adjustments made:

- Resolved its two open TODOs: components are forbidden on `kind: 'annotation'` (not merely unused), and marks MAY be component-less (they then render as class-only wrappers `mark-<type>` — this keeps CSS-only styling working, consistent with the current `anno-*` mechanism from commit 4553059).
- Added concrete pieces the plan omitted: `available_mark_types` split in Session, Node.svelte class generation, paste_utils/transforms handling, `get_node_array_annotations` split, types.d.ts, README, and the deletion of `validate_in_place_annotations_do_not_overlap`.
- Names endorsed: `marks`/`annotations` (ProseMirror precedent for marks; "annotations" matches comment/overlay usage). Neutral range type named `NodeRange` (avoid bare `Range` — clashes with DOM `Range`).

## Verification

- `npm run build` and `npm run dev`; in the demo: toggle bold/italic/highlight/link on text, toggle section and marker on the body node array; verify marker can overlap a section, section toggle removes touched sections, undo/redo works, copy/paste preserves both marks and markers.
- User runs `npm run test:unit` / `npm run test:e2e` manually (do not run automatically).

johannesmutter:
Great plan and much clearer mental model. Also this is now the right cut: keying exclusivity off `kind` instead of component registration removes the fragility (rendering config no longer changes document semantics). 

A couple of thoughts:

The mark / annotation split reads as a list of behaviors and there's a principle underneath (which should be documented in the Readme) that justifies all of them:

**Marks are semantic roles, not composable paint**: roles don't stack (a span is a link, or strong, or emphasized; "strong-emphasis" is a *fourth role you name*, never two you compose). 

**Annotations overlap because metadata is inherently multi-dimensional**: a span can be a comment AND AI-authored AND flagged for review at once.

I like that it turns "marks are exclusive" from a limitation into a stance. And when someone asks about "link + bold" mark … "this link is bold" it should be a *variant of the link role* (a `link` layout: underlined / bold / button), not a second mark. Avoiding the mess you see in other editors… 

---

**The one gap: things without a hitbox.** Rendered marks are DOM elements you can click. But annotations (and component-less marks) have no hitbox and on text specifically, since only marks fragment, a data-only annotation has *no element at all* … That leads to 2x questions:

1. How is a text annotation rendered/interacted with, given it never fragments? (CSS Custom Highlight API works for paint but is not **native** hit-testable, props-only means invisible until the app draws it.)
2. Where do view-only decorations live: spelling squiggles, search hits, collab carets, AI-suggestion previews? These are *derived*, not document data; they shouldn't sit in the persisted `annotations` array (undo/serialization pollution).

I think both could resolve with a single primitive svedit could expose:

**`range → reactive anchor/rect`** — given `(path, start_offset, end_offset)`, return a screen rect / CSS anchor that stays attached as the doc changes (or just zero width invisible non editable start and end reference rects, probably more flexible for text)

That hook serves every elementless case at once: annotation affordances (a margin comment button), component-less marks, and all ephemeral decorations.  With the app owning the decoration state and svedit only lending the position mapping it already does for selection, through the same anchor-positioning we use for node gaps and `Overlays.svelte`. 

So that means two layers (marks and annotations) and one rendering hook.

---

(Separately, and *not* for this PR: range *anchoring*: integer offsets vs. node-id anchors for node arrays (see per-character token IDs, e.g. Yjs or [Peritext](https://www.inkandswitch.com/peritext/)).

johannesmutter:
If you run into Fable limits, let me know 😄

michael:
Yeah, I feel so too. And why this is a bit of a bigger change, it's good to do it now, and have no breaking changes (from the data-model side) in the future.

Yes, marks are straight-forward now, with the exclusivity rule. Rendering annotations on text is indeed tricky. Maybe the start/end anchors approach that you suggested a while ago, can help us here. Would like to make a cut now though to finalize the release, but throw-away experimentation in a branch is welcome, and needed to get a clearer picture.

PR (WIP) is here: #321 

_____________

Regarding Fable: Some things I had in mind to use it for today+tomorrow:

- Some planning work regarding how a lightweight plugin/package infrastructure for Svedit could look like. Got the feedback of a Svedit dev, that they have troubles adding new nodes etc. without forgetting stuff like a missing inserter, updating the layouts hash etc. I do think that plugin infra should be optional, so you can stay close to the metal if you want to but for apps it makes sense to group functionality and enable/disable it. E.g. to have a Gallery plugin, that then comes with a set of nodes etc. (will probably do this today or tomorrow, just to have some insights, not to implement it yet, and definitely not for the current release)
- Finding bugs: Occasionally some node gaps are not clickable, when you save and enable edit mode again they come back. Wasn't able to reproduce reliably but maybe Fable has ideas. 
- Performance bottleneck discovery: Experiment with giving Fable the task to optimize Svedit performance (so larger docs are possible), but only through eliminating costly operations if they are discovered, not to introduce any derived state and caching (because those always backfire, and usually just make some operations faster but don't work for the general case, e.g. you had a MutationObserver in node_visibility that would not reconcile when only a text property changed. but the issue remains if you make a lot of structural changes after another, like holding enter to create new text nodes, this should be fast too; now it always reconciles on each document update). So root issue performance bottleneck discovery and fixing could be a task for Fable but making sure it does so without caching.

First one, i'd do later today it hink, but on the bugs+performance side, you could do some Fable experiments. Just make sure to scope them to smaller tasks / PR, so we can cherry-pick what works and dismiss what regresses or doesn't feel ready.

michael:
Commits are included in https://github.com/michael/svedit/pull/321

johannesmutter:
> * Performance bottleneck discovery

I was letting Fable run  performance optimisation loops all night, got some promising results. Will do a PR shortly. 

Screenshot 2026-07-07 003327 Claude@2x

PR #324, opened 2026-07-06
Fix node gaps becoming unclickable after DOM recreation (dev-mode HMR)

michael:
## Problem

Node gaps occasionally became unclickable in dev mode after saving a code change, and only recovered after the next document edit.

Root cause: the `.positioned` class (which activates anchor positioning and makes a gap's hit area clickable) was toggled **imperatively** on gap elements by the visibility registry, and fully resynced only when `session.doc` or `editable` changed. When Svelte recreates DOM without a document change — which is exactly what a Vite/Svelte HMR component swap does — the fresh `.node-gap` elements never received the class, leaving them 0×0 with `pointer-events: none`. The next document edit triggered `reconcile()` and everything snapped back, matching the observed "edit again and it works" behavior.

## Reproduction

New test in `gap_visibility.svelte.test.js` ("DOM recreation without document change") simulates HMR by swapping `session.config.node_components.story` for an identical wrapper component (`StoryHmrProxy`), which remounts the story subtree — fresh node and gap DOM — with no document change. It failed before the fix (recreated gaps lacked `.positioned`) and passes after.

## Fix

`NodeGap` now derives `positioned` from the registry's `near_map`/`edge_map` via `$derived` and applies it declaratively with `class:positioned` — the same pattern `NodeGapMarkers` already uses (and why the markers survived HMR while the gaps broke). Since Svelte owns the class, it is re-applied automatically on any DOM recreation.

The now-redundant imperative path is deleted (~60 lines): `sync_gap_class`, `sync_gaps_around_node`, `sync_array_edge_gaps`, `#sync_first_last_gaps`, the gap-resync loop in `reconcile()`, and `GAP_SELECTOR`. The scroll handler and IO callback now just write to the reactive maps; `should_position_gap` is exported and no longer wrapped in `untrack()` so per-key subscriptions form.

## Performance

Unchanged. `SvelteMap`/`SvelteSet` track reads per key, so an IO write for one node re-evaluates only the two adjacent gaps — same O(1) update as the old manual sibling traversal. Off-screen gaps remain un-positioned with zero anchor-resolution cost.

## Reviewer notes

- Verified: all 15 gap-visibility tests (incl. the new repro), all 6 selection-scroll tests, `npm run check` clean.
- Known follow-up (out of scope): the IntersectionObserver still tracks *elements*, so nodes recreated by HMR are unobserved until the next document change — `near_map` can go stale on scroll in that window. Planned fix is attachment-based node registration (`{@attach}`) so observe/unobserve follows element lifecycle.

### DIFF
--- a/src/lib/Node.svelte
+++ b/src/lib/Node.svelte
@@ -54,6 +54,7 @@
 	data-type="node"
 	style="anchor-name: --{path_str};{style}"
 	{...rest}
+	{@attach svedit.visibility_registry.track_node(path_str)}
 >
 	{@render children()}
 </svelte:element>
--- a/src/lib/NodeArrayProperty.svelte
+++ b/src/lib/NodeArrayProperty.svelte
@@ -132,6 +132,7 @@
 	data-path={path_str}
 	style="anchor-name: --{path_str};{style ? ` ${style}` : ''}"
 	{...rest}
+	{@attach svedit.visibility_registry.track_array(path_str)}
 >
 	{#if node_ids.length === 0 && svedit.editable}
 		<div
@@ -139,6 +140,7 @@
 			data-path={serialize_path([...path, 0])}
 			data-type="node"
 			style="anchor-name: --{serialize_path([...path, 0])};"
+			{@attach svedit.visibility_registry.track_node(serialize_path([...path, 0]))}
 		>
 			<NodeGap array_path={path} offset={0} count={0} empty />
 		</div>
--- a/src/lib/NodeGap.svelte
+++ b/src/lib/NodeGap.svelte
@@ -13,6 +13,7 @@
 <script>
 	import { serialize_path } from './utils.js';
 	import { getContext } from 'svelte';
+	import { should_position_gap } from './node_visibility.svelte.js';
 
 	/**
 	 * ┌─────────────────────────────────────────────────────────────────┐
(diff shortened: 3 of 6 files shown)


michael:
~~**Known limitation (follow-up candidate): `edge_map` can go stale on window resize.**~~

**Resolved in this PR** (839f940): a shared `ResizeObserver` on registered arrays now re-syncs `edge_map` whenever an array's box changes size — covering window resizes, sidebar toggles, responsive breakpoints and late-loading content uniformly, not just window resize.

Context for reviewers: `edge_map` (the per-array `{first, last}` scroll-position state that gates edge gaps) was previously recomputed only on array mount, on scroll inside the array, and after document changes. A box-size change can flip an array between fitting and overflowing while `scrollLeft` stays 0 — none of those triggers fire, so edge gaps kept their pre-resize visibility until the next scroll or edit.

The accompanying test is deliberately built around a **partial** clip of the trailing node: a full clip zeroes the node's IntersectionObserver intersection and hides the gap via `near_map`, which would make the test pass even without the ResizeObserver. With a partial clip no IO threshold is crossed, so only the resize-driven `edge_map` sync can hide the gap. Verified the test fails with the observer disabled.

michael:
@johannesmutter when you get some time, can you review this one first (if anything looks off or regresses)? 

It solves the 'node-gaps-turn unclickable during HMR' issue for me.

If all is good with this PR, check #325 too? (It's a simplification, reducing source state - same behavior.)

michael:
Covered via https://github.com/michael/svedit/pull/329

PR #325, opened 2026-07-06
Simplify: collapse near_map into array_indices as the single near-state store

michael:
**Stacked on #324** — review that one first; this is a pure refactor on top (no behavior change intended). If both are approved: merge this into #324's branch, then merge #324 into main.

## What

The visibility registry kept node nearness in two stores written together at every site: `near_map` (full path → `true`, read by NodeGap) and the per-array `array_indices` SvelteSets (read by NodeGapMarkers). Every `near_map` read had the form "array path + child index", so both consumers can read the same per-array sets. `near_map` is deleted; `array_indices` is the single source of truth.

This is the same disease #324 treats — two copies of one fact that must be kept in sync — applied to the registry's own internals. It also drops dead state: `near_map` stored entries for nodes that aren't node-array children (document root), which nothing ever read.

## The interesting part: a real Svelte reactivity trap

The naive version of this refactor (calling `registry.get_array_indices(path).has(i)` inside `should_position_gap`) **silently broke the first gap of every array**, caught by the existing test suite. Root cause: `get_array_indices` lazily creates the `SvelteSet`, and Svelte intentionally does not track dependencies on state created inside the currently-evaluating reaction. So the first consumer to touch an array (always the `gap-before`, first in DOM order) created the set inside its own `$derived`, never subscribed, and froze un-positioned forever — while every later gap read the existing set and worked. `NodeGapMarkers` only ever worked because it happens to hold the set in one `$derived` and read it in another.

The fix makes that split explicit and documented:

- `NodeGap` acquires the set in its own `$derived` (`near_indices`) and reads membership in the `positioned` derived — mirroring `NodeGapMarkers`.
- `should_position_gap` is now a **pure function** `(near, edge_state, offset, is_last, empty)` — no registry access, trivially testable.
- `get_array_indices` carries a doc warning that the two-derived split is load-bearing.

## Verified

All suites pass: gap_visibility (16, incl. the HMR repro and the resize test), selection_scroll (6), keymap_continuity (4), Svedit/Session/annotations (93). `npm run check` clean.

### DIFF
--- a/src/lib/NodeGap.svelte
+++ b/src/lib/NodeGap.svelte
@@ -39,8 +39,9 @@
 	 * Always present in the DOM when editable (stable structure for
 	 * selection anchoring, scrollTo, etc.). Anchor positioning is
 	 * activated lazily via the `.positioned` class, derived reactively
-	 * from the visibility registry's near_map / edge_map. SvelteMap
-	 * tracks reads at per-key granularity, so a registry write only
+	 * from the visibility registry's array_indices / edge_map. Svelte's
+	 * reactive collections track reads at per-key granularity, so a
+	 * registry write only
 	 * re-evaluates the adjacent gaps. Applying the class declaratively
 	 * (instead of imperative classList toggles) means it survives DOM
 	 * recreation without a document change (e.g. dev-mode HMR).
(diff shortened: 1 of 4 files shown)


michael:
Covered via https://github.com/michael/svedit/pull/329

PR #326, opened 2026-07-07
Eliminate O(N) layout and data-layer costs for large documents

johannesmutter:
Makes large documents (500–5000+ nodes) fast by **eliminating** costly per-operation work — no caches or derived state added, only better algorithms and fewer forced layouts.

## Root cause

The dominant cost wasn't JS — it was **layout**. Every off-screen `NodeGap` kept a zero-size `position:absolute` selectable, and the browser re-lays-out every out-of-flow box on every layout pass (~430ms per pass at 2000 nodes, ~3500 gaps). Since the per-change `reconcile()` reads a rect for every node and selection rendering forces layout, this throttled typing, Enter, resize, and scroll alike. On top of that, the data layer did full-document scans/clones per operation.

Found and measured with a small perftest lab (`/perftest`, `window.__bench`) added here.

## Changes

**Layout**
- `NodeGap`: unpositioned non-empty gap selectables are `display:none` — off-screen gaps contribute no layout box (~430ms → ~23ms per full layout at 2000 nodes). DOM structure and range read-back are preserved; empty-array gaps excluded.
- `Svedit`: drop redundant `removeAllRanges()` before `setBaseAndExtent` (it already replaces the selection; each selection call forces layout).
- `node_visibility.sync_edge_state`: check computed overflow before reading scroll metrics so non-clipping arrays don't force layout.
- `TextProperty`: fast path for unannotated text (skips two `Intl.Segmenter` passes per property per change); length-based `is_empty`.

**Data layer**
- `doc_utils`: `create_document_draft` + `apply_op_to_draft` — one nodes-map copy per transaction/undo/redo with node-level copy-on-write, replacing the per-op full-map copy. New `visit_node_references` / `build_reference_counts`.
- `Transaction`: draft-based apply; Set-based removed-id diff (was O(N²)); single-scan cascade GC with incremental decrement (was a full scan per candidate); `changed_node_types` flag; `traverse_ids` for reference walks.
- `Session`: `validate_transaction_result` does at most one referrer scan (only on delete or type change); draft-based undo/redo; `traverse` no longer snapshots the whole doc; ops-less apply keeps the doc reference.
- `utils`: `traverse_ids` (id-only, clone-free traversal).

All reference kinds of the current model are handled: `node`, `node_array.nodes`, and mark **and** annotation `node_id`s on both `text` and `node_array`.

## Results (Chrome, dev)

| Operation | Before | After |
|---|---|---|
| Copy 200 nodes (2000-doc) | 2420 ms | **2 ms** |
| Delete 200 nodes, data (2000) | 604 ms | **20 ms** |
| Enter render (500 / 1000 / 2000) | 88 / 329 / 1615 ms | **51 / 131 / 374 ms** |
| Type render (2000) | 1324 ms | **123 ms** |
| Full relayout / resize step (2000) | 438 ms | **23 ms** |

Residual per-keystroke cost is the unavoidable `setBaseAndExtent` layout — the browser's floor for this DOM size.

## Verification

21 in-browser correctness checks pass, including mark-node and annotation-node cascade deletion, undo/redo of cascades, split/join, and validation still rejecting invalid input. Production build and lint are clean.

## Notes for review

- Off-screen gaps are no longer native arrow-key stops (they previously rendered an *invisible* caret with no marker, so arguably a fix) — worth a feel-check on horizontally-scrolling arrays.
- Docs passed to `Session` must be plain objects (they are in this repo); `traverse`/copy would throw on a `$state`-proxied doc now that the per-call full-doc snapshot is gone.
- Scroll/resize **FPS** wasn't measured here (the lab ran in a headless tab); a quick check in a visible browser at 1000+ nodes is worthwhile.
- `src/routes/perftest/*` + `src/hooks.server.js` are the measurement lab; drop them from the merge if you'd rather not ship the harness.

### DIFF
--- a/src/hooks.server.js
+++ b/src/hooks.server.js
@@ -1,0 +1,10 @@
+/**
+ * Enables the JS self-profiling API (used by the perftest lab) on dev-served
+ * pages. Vite's server.headers doesn't apply to SvelteKit-rendered HTML, so
+ * the header is set here.
+ */
+export async function handle({ event, resolve }) {
+	const response = await resolve(event);
+	response.headers.set('Document-Policy', 'js-profiling');
+	return response;
+}
--- a/src/lib/NodeGap.svelte
+++ b/src/lib/NodeGap.svelte
@@ -111,8 +111,29 @@
 	}
 
 	/* ------------------------------------------------------------------ */
-	/* Un-positioned: minimal in-DOM presence, no anchor resolution        */
+	/* Un-positioned: no layout box at all                                 */
 	/* ------------------------------------------------------------------ */
+
+	/*
+	 * display: none is load-bearing for large documents. A zero-size
+	 * absolutely-positioned selectable still belongs to the containing
+	 * block's out-of-flow list, and the browser lays out EVERY such box
+	 * on EVERY layout pass — measured ~430ms per pass at 2000 nodes
+	 * (~3500 gaps) in Chrome vs ~23ms with the boxes removed. That cost
+	 * hits every keystroke (the per-change reconcile reads a rect for
+	 * every node, and selection rendering forces layout), every window
+	 * resize frame, and every scroll-triggered layout.
+	 *
+	 * Off-screen gaps therefore contribute no layout box. DOM structure
+	 * stays stable (the .node-gap wrapper and selectable elements remain),
+	 * and DOM Ranges may still point into display:none elements, so
+	 * programmatic node selections targeting off-screen gaps keep working.
+	 * Empty-array gaps are excluded: they must stay clickable/visible even
+	 * before reconcile positions them.
+	 */
+	:global(.node-gap:not(.positioned):not(.empty) .svedit-selectable) {
+		display: none;
+	}
 
 	:global(.node-gap:not(.positioned) .svedit-selectable) {
 		position: absolute;
--- a/src/lib/Session.svelte.js
+++ b/src/lib/Session.svelte.js
@@ -1,11 +1,12 @@
 import Transaction from './Transaction.svelte.js';
-import { char_slice, traverse } from './utils.js';
+import { char_slice, traverse, traverse_ids } from './utils.js';
 import {
 	get as doc_get,
 	property_type as doc_property_type,
 	kind as doc_kind,
 	inspect as doc_inspect,
-	apply_op,
+	create_document_draft,
+	apply_op_to_draft,
 	count_references as doc_count_references,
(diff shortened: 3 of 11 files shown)


michael:
Okay wow. That's a crazy boost. Will do testing throughout the day.

Does this also address the things from here #324 and #325? Or focus only on performance. If the latter, I'm wondering in which order we are trying to land the fixes and performance improvements. #324 contains a test revealing the issue I had.

michael:
You could also try to port the work from #324 and #325 on this (new PR with base perf-optimizations-2026-07-07), with your own angle and considerations, then we can test the whole thing together.

But step-by-step is also fine. I don't know what makes more sense. Letting you and Fable this today, and I'm here for testing and feedback.

johannesmutter:
This one is only performance. Also just ran out of my weekly usage limits :D 
I'll have a look first at those other two PRs #324 and #325 now

michael:
I've been testing quite a bit, and seen no problems. What I don't yet fully understand is the draft-based apply. It seems to be safe, but I want to make sure the copy-on-write principle is not violated, leaving potential footguns where doc is not immutable anymore. I wonder if that particular optimization has much impact, as it seems the real cost was node gaps trigger layout-reflows.

johannesmutter:
The draft-based apply has quite an impact, too:

Doc size | baseline | with draft
-- | -- | --
1000 nodes | 56.6 ms / 88 ms | 0.9 ms / 17 ms
2000 nodes | 109.1 ms / 139 ms | 0.9 ms / 28 ms

> baseline _apply_op scales with document size (57ms → 109ms as N doubles) because every one of the 201 ops re-copies the whole nodes map — that's the O(N·M). With the draft it's flat at 0.9ms regardless of N (one map copy per transaction). It was ~65–78% of the paste time and is now noise. For delete/cut it stacks with the single-scan cascade fix (that's why bulk delete went 604ms → 20ms — neither fix alone gets you there).

michael:
Good to know. Thanks! It's an amazing performance win, just need a bit of time to digest this. Don't want to merge blindly. Weird times, it's tempting to just merge the thing that apparently works and is tested, but I want to be able to fully understand the code myself, so i can challenge it.

However I think the plan is to land this first (and be sure it doesn't regress), and then address #324 and #325 next. If you can take a look at those in the meanwhile and test and let me know if something stinks from your view that would be helpful.

michael:
Landed via https://github.com/michael/svedit/pull/329

PR #328, opened 2026-07-07
Guard against null focusNode in deferred scroll-into-view

johannesmutter:
**What happens**

`render_selection` sets the DOM selection, then scrolls it into view a tick later:

```js
setTimeout(() => {
	const selectedElement = dom_selection.focusNode.parentElement; // throws
	...
}, 0);
```

If focus leaves the canvas before that timeout fires (e.g. into a form `<input>` in a toolbar or dialog), the browser clears the document selection, so `dom_selection.focusNode` is `null` and `.parentElement` throws:

```
Uncaught TypeError: Cannot read properties of null (reading 'parentElement')
```

**Not a DOM-stability issue.** Logged at the throw site: the intended `focus_node` is still `isConnected`, `getSelection().type === 'None'`, and `activeElement` is the input. Nothing re-rendered out from under the selection; the selection just moved to the input.

**Repro:** select text, focus an `<input>` outside the canvas (a link editor, etc.), then click back into the text.

**Fix:** optional-chain the `focusNode` read. A deferred read of the live `Selection` can't assume `focusNode` still exists, and if the selection has left the canvas there is nothing to scroll to.

### DIFF
--- a/src/lib/Svedit.svelte
+++ b/src/lib/Svedit.svelte
@@ -1505,9 +1505,12 @@ ${fallback_html}`;
 				focus_node_offset
 			);
 
-			// Scroll the selection into view
+			// Scroll the selection into view. Deferred a tick, so focus may have
+			// left the canvas by now (e.g. into a form input or dialog), which
+			// clears the document selection and leaves focusNode null. Nothing to
+			// scroll to in that case.
 			setTimeout(() => {
-				const selectedElement = dom_selection.focusNode.parentElement;
+				const selectedElement = dom_selection.focusNode?.parentElement;
 				if (selectedElement) {
 					selectedElement.scrollIntoView({ block: 'nearest', inline: 'nearest' });
 				}


michael:
> In an editor whose mark component re-renders its text node: apply the mark over a range, then click another text position. The click sets a new selection, the mark component re-renders and swaps its text node

Can't follow. Why does this mark component rerender? Is that something app-specific? Can it be reproduced in the current Svedit demo?

Don't fully understand yet, but hints the root issue is somewhere else. Feels a bit like this situation we had when the selection-fragment rendering during dragging a selection was changing the DOM structure underneath destroying the selection. This must not happend. Markup needs to be stable during selection drag etc.

But I'm just doing guessing here, don't really understand the context.

johannesmutter:
The PR description by Claude was wrong. Updated it.

> Is it app-specific? 

The difference is my annotation editor does two things in one flow: it toggles the link mark, then focuses its URL input. Toggling the mark changes the selection, which makes svedit schedule its deferred scrollIntoView. Then we focus the input, which clears the selection. The setTimeout(0) fires after that and reads focusNode = null.

So the trigger is "change the selection, then move focus to an input in the same tick." e.g. your editable-website doesn't change the selection right before focusing its link input, so there's no pending scroll to race.

I could reorder my side (focus the input before toggling), but svedit reading the live selection in a setTimeout still can't assume focusNode is there. The guard covers that.

> Reproducible in the demo? 

No. The demo's link editor is a blocking window.prompt, so it never does that sequence.

**On the DOM-stability hunch:** not this one. The markup is stable (the node stays connected); the selection is lost to a focus change, not a re-render. Different from the drag case. The guard just makes svedit's deferred read tolerate a selection that moved.

michael:
Added a test for this and then fixed with a different approach: Capture the reference before the deferred block, so we can be sure the reference is there, if even the selection moved elsewhere.

Can you check https://github.com/michael/svedit/pull/330 and let me know if this fix works for you?

michael:
Fixed via #330

PR #331, opened 2026-07-08
Stabilize deferred node-selection scrolling

michael:
Capture node-selection scroll intent before the deferred scroll callback runs, instead of recomputing model state from the live session inside the timeout.

Keep viewport intersection checks in the deferred callback so layout is measured after Svelte/browser updates have settled.


michael:
@johannesmutter this is similar hardening to #330. Just for node selections. If you have time, maybe you can do a quick scan + testing for possible regressions. But should be more robust now.

Issue #335, opened 2026-07-16
svedit as an op-emitting engine


johannesmutter:
**Status:** draft for discussion
**Scope:** Session / Transaction internals. All changes additive; no breaking API changes intended.
**Shape:** one RFC to agree on direction, then four issues, each landing as a series of small, independently reviewable PRs.

## Problem

svedit treats persistence as the host's problem and gives the host exactly one tool: serialize the whole document (`session.to_json()`) at a moment the app chooses ("save"). Three pressures make that too coarse for the apps svedit is being used in:

**1. Users expect save-less editing.** Crash safety and continuous persistence (the Apple Notes/iA Writer/Notion/Docs/Figma baseline) require the host to observe every committed change the moment it happens. Today there is no change notification at all. Worse, `undo()` and `redo()` mutate `session.doc` directly, bypassing `apply()`, so even a host that wraps `apply` desyncs from the real document state on the first undo.

**2. External writers are now normal.** A) Agents edit documents headlessly. b) A second tab edits the same document. c) A sync layer or a version-restore feature wants to apply changes into a *live* session … There is no supported entry point for any of these: mutating the doc out-of-band skips validation, breaks selection, and corrupts undo bookkeeping, so hosts fall back to full reloads.

**3. Long sessions have an unbounded memory profile, and ops carry no meaning.** Every text keystroke records the *whole* property value twice: `['set', path, whole_value]` in `ops` and the previous whole value in `inverse_ops`. Typing 30 characters into a 5 KB paragraph stores roughly 300 KB of strings in history; an afternoon of writing grows without bound (batching concatenates the copies, it does not avoid them). And because every text edit is "content changed", the op stream is useless as an edit trace: no meaningful diffs, previews, or audit.

svedit is unusually well positioned to fix all three, because the architecture is already right:

- transactions commit by swapping an immutable snapshot (0.12's `create_document_draft` / `apply_op_to_draft`),
- ops are three serializable, invertible primitives (`set`, `create`, `delete`) with inverses recorded,
- `apply()` is a single choke point (except for the undo/redo bypass),
- zero runtime dependencies, so none of this drags in a framework.

The gaps are small and precise. Closing them turns svedit into an engine that *emits* and *accepts* operations, while persistence, sync, and merge policy stay entirely in host land.

## Proposal

Four parts, in dependency order.

### 1. One commit path + change notifications

[...]

PR #348, opened 2026-07-18
Add history spill, op replay and node-scoped history filtering

johannesmutter:
Part 4/4 of the op-engine series (#335). Stacked on `op-engine/3-ingest`.

- `history_limit` + `on_history_spill` session options: bounded in-memory history for long sessions; spilled entries are plain JSON for the host to persist; undo depth is bounded by the retained window
- `apply_ops(doc, ops)`: the replay primitive (materialize any version from a snapshot plus recorded ops); undo/redo/ingest now share it, so live sessions and host replay use one code path
- `session.history_for(node_ids)` + `op_target_node_id(op)`: per-node history filtering — the base for per-node time travel ("slide through this paragraph's versions") and scoped undo, instead of global undo that teleports across the document

`apply_ops`, `op_target_node_id` and the `HistoryEntry` type are exported from the package index.

Tests: 5 for this PR. Note for `npm test` on any branch (including main): the pre-existing keymap_continuity tests are order-sensitive — `npx vitest --run src/test/annotations.svelte.test.ts src/test/keymap_continuity.svelte.test.ts` fails 3 tests on unmodified main while the file alone passes; whether a full run goes red depends on vitest's file ordering. Worth a separate issue.

### DIFF
--- a/src/lib/Session.svelte.ts
+++ b/src/lib/Session.svelte.ts
@@ -5,8 +5,8 @@ import {
 	property_type as doc_property_type,
 	kind as doc_kind,
 	inspect as doc_inspect,
-	create_document_draft,
-	apply_op_to_draft,
+	apply_ops,
+	op_target_node_id,
 	count_references as doc_count_references,
 	validate_document_schema,
 	validate_document,
@@ -57,7 +57,7 @@ function transform_offset_for_splice(offset: number, op: DocumentOperation): num
 	return start;
 }
 
-type HistoryEntry = {
+export type HistoryEntry = {
 	ops: Transaction['ops'];
 	inverse_ops: Transaction['inverse_ops'];
 	selection_before: Selection | null;
@@ -176,6 +176,11 @@ export default class Session<S extends DocumentSchema = DocumentSchema> {
 	// state: notification happens explicitly in _commit().
 	#change_listeners: Array<(change: ChangeEvent) => void> = [];
 
+	// History spill configuration (see constructor options). Plain fields —
+	// may be reconfigured at runtime.
+	history_limit = Infinity;
+	on_history_spill: ((entries: HistoryEntry[]) => void) | null = null;
+
 	// Reactive helpers for UI state
 	can_undo = $derived(this.history_index >= 0);
 	can_redo = $derived(this.history_index < this.history.length - 1);
@@ -203,7 +208,21 @@ export default class Session<S extends DocumentSchema = DocumentSchema> {
 		schema: S,
 		doc: Document,
 		config: SessionConfig,
-		options: { selection?: Selection | null } = {}
+		options: {
+			selection?: Selection | null;
+			/**
+			 * Maximum number of history entries kept in memory (default
+			 * Infinity). When exceeded, the oldest entries are handed to
+			 * on_history_spill and dropped; undo depth is bounded accordingly.
+			 */
+			history_limit?: number;
+			/**
+			 * Called with the oldest history entries when history_limit trims
+			 * them. Entries are plain JSON; persist them to replay later (see
+			 * apply_ops in doc_utils.ts).
+			 */
+			on_history_spill?: (entries: HistoryEntry[]) => void;
+		} = {}
 	) {
 		// Validate the schema first
 		validate_document_schema(schema);
@@ -216,6 +235,9 @@ export default class Session<S extends DocumentSchema = DocumentSchema> {
 
 		// Set selection after doc is initialized so validation can work properly
 		this.selection = options.selection ?? null;
+
+		this.history_limit = options.history_limit ?? Infinity;
+		this.on_history_spill = options.on_history_spill ?? null;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


michael:
What is the use-case for this? If I understand correctly, you want to be able to undo/redo changes scoped to a specific section, not the whole document?

Basically scoped undo/redo buttons? This seems quiet a bit app-specific. Like I can't imagine an app I'd build to use this. I also see consistency challenges. Scoped by node is not enough I believe because you are likely often dealing with a whole subtree of the document. Like say you have a list with list items, then you can not just scope history entries to that list, but also all referenced elements. If you ever want to reuse nodes in different place (=graph not just a tree) then things get even more involved.

michael:
> Pass a node id plus its owned subgraph ids (see get_referenced_nodes) to scope to a node including its mark and annotation nodes.

Ah I see you considered this. Just not a fan of that API, seems brittle. Okay in app-land, if you know your domain model (e.g. are sure you are dealing with a tree, not a graph).

How about just extending Session in app-land for this filtering thing?

michael:
Ran app-land filtering idea through GPT 5.6, just for additional context/ideas.

______________

The PR does not actually implement scoped undo/redo. `history_for(nodeIds)` only returns matching global history entries. The PR describes two possible consumers: per-node version browsing and a foundation for scoped undo. [[PR #348](https://github.com/michael/svedit/pull/348)](https://github.com/michael/svedit/pull/348)

I share your concern. Filtering is easy; defining what constitutes a coherent “node scope” is the hard, application-specific part.

A few complications:

- A history entry is an atomic transaction and may modify multiple sections. Matching one node does not mean the entry can safely be partially undone.
- The scope is usually a logical aggregate, not a node: section → blocks → list items → marks/annotations.
- `get_referenced_nodes()` follows the current graph. The relevant subgraph may have been different when an old history entry was created.
- Shared nodes need an explicit policy: should their edits appear in every referencing section, one owning section, or only global history?
- Undoing an older matching entry out of order can invalidate later operations.

For filtering alone, extending `Session`—or even using a standalone app utility—is sufficient:

```ts
import {
  Session,
  type DocumentOperation,
  type NodeId
} from 'svedit';

function targetNodeId(op: DocumentOperation): NodeId | undefined {
  switch (op[0]) {
    case 'set':
    case 'splice':
      return op[1][0];
    case 'create':
      return op[1].id;
    case 'delete':
      return op[1];
  }
}

class MyAppSession extends Session {
  historyForSection(sectionId: NodeId) {
    const scope = this.resolveSectionScope(sectionId);

    return this.history
      .map((entry, index) => ({ entry, index }))
      .filter(({ entry }) =>
        entry.ops.some((op) => {
          const id = targetNodeId(op);
          return id !== undefined && scope.has(id);
        })
      );
  }

  private resolveSectionScope(sectionId: NodeId): Set<NodeId> {
    const referenced = this.get_referenced_nodes(sectionId);

    // This is where the app expresses its ownership rules:
    // include descendants, decorations, perhaps exclude shared content, etc.
    return new Set([sectionId, ...referenced]);
  }
}
```

That reproduces `history_for()` without a core change, while making it clear that `resolveSectionScope()` belongs to the application’s domain model.

For actual “undo this section,” I would avoid moving the global `history_index` or selectively applying an entry’s inverse operations. I would model it as “restore section to version”:

```ts
restoreSection(sectionId: NodeId, version: number) {
  const historicalDoc = this.materializeDocumentAt(version);
  const tr = this.tr;

  reconcileSection({
    tr,
    sectionId,
    currentDoc: this.doc,
    historicalDoc
  });

  // A normal new transaction, validated and globally undoable.
  return this.apply(tr);
}
```

Here `reconcileSection()` is app code that knows ownership, shared-node behavior, and whether to clone or preserve referenced nodes. The restore becomes a new global history entry rather than a nonlinear mutation of existing history.

So my inclination for the PR would be:

- Keep `apply_ops()` and history spilling; those are generally useful persistence/replay primitives.
- Possibly keep `op_target_node_id()` as a low-level op utility, though it is trivial for apps to implement.
- Leave `history_for()` out of core until there is a compelling generic use case and well-defined semantics.
- Build per-section version browsing/restoration in app-land, preferably with scopes recorded when changes occur rather than reconstructed solely from the current graph.

The strongest argument is not merely that the method is small enough to live in the app. It is that the method’s name suggests stronger semantics than it provides: it finds entries whose operations address certain IDs; it does not produce a self-contained or safely undoable history for a node.

johannesmutter:
> What is the use-case for this? If I understand correctly, you want to be able to undo/redo changes scoped to a specific section, not the whole document? Basically scoped undo/redo buttons …

Yes scoped undo/ redo to a specific node or selection of nodes for example. Not with buttons, probably more something like a super minimal (slider) time machine/ git version history interface.

Here's the use case:
Multiple agents, myself (and other human users) editing a document. Edits happen all over the place, and not in a meaningful order. Most of the edits come from agents. Classic global undo/ redo had its place in single player. But also there it failed when I temporarily switch to another section, spend 10 minutes doing edits there and then coming back to where I started, global undo/ redo will not allow me to time travel back the changes I made to this particular section, or only with a risky maneuver: travel back to the target change, copy the old state and then redo forwards again, hopefully without making a single change to the version history, as otherwise all other future changes are gone. I have felt this problem in dozens of apps, from code, to prose to design. The least pain is e.g. in tools like Figma, because there it's quite natural to keep copies of old versions around for comparison on the canvas, so the need to time travel is solved through spatial means.

But in a text document that is heavily edited, where you try to develop a storyline and make changes here and there, there's no such escape hatch (other than duplicating the document).
When editing with LLMs I find it gets messy very quickly, and I feel no oversight about what changed and how to maybe restore surgically. 

---

The point of this PR is not to implement scoped undo/redo but to lay the foundation for apps to implement it. Though maybe it can be done a bit more minimally.

johannesmutter:
A personal motivation for me is to have a version history persisted per doc/ node, not just in a session. 
So think of a document as a history made up of "commits". Each edit made in a document means an addition to the history of the document.

johannesmutter:
> For actual “undo this section,” I would avoid moving the global history_index or selectively applying an entry’s inverse operations. I would model it as “restore section to version” … The restore becomes a new global history entry rather than a nonlinear mutation of existing history.

Those a very good comments

PR #350, opened 2026-07-21
Don't scroll annotations into view that are already in view

michael:


michael:
Fixed via https://github.com/michael/svedit/pull/354

PR #357, opened 2026-07-23
Position floating insert tool using node gap orientation

michael:
Not sure if container style queries are allowed cross-browser already? Also doesn't place the tool "before the first gap" yet.

### DIFF
--- a/src/lib/NodeGapMarkers.svelte
+++ b/src/lib/NodeGapMarkers.svelte
@@ -156,7 +156,7 @@
 		class:first={gap.is_first}
 		class:last={gap.is_last}
 		class:pair={gap.has_pair}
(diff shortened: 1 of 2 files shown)


michael:
Also it should fallback if there's not enough space at the intended placement.

michael:
@johannesmutter maybe you can review/pick up? No rush

johannesmutter:
oh just saw this ... I don't think placement works as how it's done now:
Screenshot 2026-07-25 003748 Svedit - A tiny library for building editable websites in Svelte@2x

michael:
Yeah hard to find a good place for those in "tight horizontals". Let's get rid of the pluses.

PR #358, opened 2026-07-25
Node Gap Tools: a snippet for custom tools that should render in the node gaps

johannesmutter:
It becomes quite complicated trying to render custom tools in or near node gaps, because they are not part of the hierarchy (usually they are top level), so straightforward css inheritance e.g. of `--row:1 or 0` for placement optimisation is not possible.

This PR changes that by making it possible to configure a component via `node_gap_tools` that renders in the snippet inside NodeGap.

4x more LOC in core lib.


michael:
Makes sense. Good approach with opt-in system_component.

